PR #2, opened 2022-12-01
try to fix where the assets are

b-pmcg:


package.json
```diff
@@ -40,19 +40,22 @@
...
-		"linux": {
-			"category": "Finance"
-		},
-		"dmg": {
-			"icon": false
-		},
+    ],
+    "extraResources": [
+      "public/**"
+    ],
```
b-pmcg, line 50: this copies the public files in `Contents/Resources/` folder where Macos is looking for it. Not sure if it works for windows too.

scripts/electron.js
```diff
@@ -4,8 +4,8 @@ const next = require('next');
 const { autoUpdater } = require('electron-updater');
 const path = require('path');
 
-const dockIcon = path.join(__dirname, 'public', 'images', 'logo.png');
-const trayIcon = path.join(__dirname, 'public', 'images', 'logo.png');
+const dockIcon = path.join(__dirname, '..', '..', 'public', 'images', 'logo.png');
+const trayIcon = path.join(__dirname, '..', '..', 'public', 'images', 'logo.png');
```
b-pmcg, line 8: this tells it to back up two dirs to find the files in `Contents/Resources/` on mac. This only works when building the app though. When running it locally, the file structure is different, and you only need to back it up one dir, so it looks like this

```
const dockIcon = path.join(__dirname, '..', 'public', 'images', 'logo.png');
const trayIcon = path.join(__dirname, '..', 'public', 'images', 'logo.png');
```

rafinskipg (approved):

PR #6, opened 2022-12-09
Config setup

adamgoth:


electron-app/main.js
```diff
@@ -54,6 +54,8 @@ async function launchApp() {
     win.center();
     splash.close();
     win.show();
+    // TODO remove this line or only open in dev mode
+    win.webContents.openDevTools();
```
rafinskipg, line 58: add the "if condition" of "isDev" to open or not the devtools

rafinskipg (approved):

PR #8, opened 2022-12-12
Add userconfig and siteconfig

rafinskipg:
https://www.loom.com/share/11c87cbbdf744b18ab4df4c38c18ebab


src/modules/config/context/ConfigContext.tsx
```diff
@@ -0,0 +1,128 @@
...
+// It stores all the RPCs the application will use, and also the user configured-ones
+const defaultConfig: SiteConfig = {
+  name: 'MetaDAO frontend',
+  description: 'Interact with MakerDAO features',
+  palette: {
+    text: 'black',
+    background: 'white'
+  },
+  logo: './images/logo.png',
+  favicon: './images/logo.png',
```
adamgoth, line 20: I'm not sure I understand the benefit of having both `metadao.config.json` file and the same default config values stored here too. The only way to update either one is via code so it seems like it's unnecessary to store in two places. It makes sense for the `rpcs` though.

src/modules/config/context/ConfigContext.tsx
```diff
@@ -0,0 +1,128 @@
...
+};
+
+// Default user config
+const defaultUserConfig: UserConfig = {
+  rpcs: [],
+  theme: siteConfig.theme
+};
+
+export interface ConfigContextProps {
+  config: SiteConfig;
```
adamgoth, line 59: Maybe `siteConfig` here instead of just `config` for clarity? Lots of names with config in the file.

PR #11, opened 2022-12-14
reorganize contracts

rafinskipg:


src/pages/Dsr.tsx
```diff
@@ -1,20 +1,25 @@
 import React from 'react';
 import { Layout } from '../modules/layout/components/Layout';
 import { Heading, Box, Label, Text } from 'theme-ui';
-import { useNetwork } from '../modules/web3/hooks/useNetwork';
-import { useDsrTotalSupply } from '../modules/web3/hooks/useDsrTotalSupply';
+import { useContractRead } from '../modules/web3/hooks/useContractRead';
+import { chainId } from 'wagmi';
+import LoadingPlaceholder from '../modules/ui/components/LoadingPlaceholder';
+import { formatUnits } from 'ethers/lib/utils.js';
+import { BigNumberish } from 'ethers';
```
adamgoth, line 8: These are all the imports that I think it'd be nice to not have to do on every page file. It puts a bunch of logic into the UI code.

But if we're going to go this route then we can delete the useContractRead hook.

PR #15, opened 2023-01-06
Monorepo: initial setup

rafinskipg:
Adds monorepo:
- managed with PNPM workspaces:

Packages: 
- apps/webapp -> Web app that serves as interface
- packages/hooks -> Subdao / maker core features hooks with wagmi 
- packages/contracts -> Static contracts mapping for the maker protocol

To run the webapp you need to check the README.md instructions

- [x] Includes Readme for the whole project and packages
- [x] Includes "build / lint / test" workflow

In next PRS i'll include:
- [ ] Includes "Release" workflow and documentation around releasing
- [ ] The app uses the inner packages
- [ ] There is documentation on how to run / fork the webapp


adamgoth (commented):
Assuming we'll want to make use of packages like this, I think this looks like a good setup. Nice work. 

I noticed the README talks about using the other packages, such as hooks, but it doesn't look like the PR goes as far as importing the package anywhere in the app (from what I can tell).

Can we maybe add an example of that?

rafinskipg:
> Assuming we'll want to make use of packages like this, I think this looks like a good setup. Nice work.
> 
> I noticed the README talks about using the other packages, such as hooks, but it doesn't look like the PR goes as far as importing the package anywhere in the app (from what I can tell).
> 
> Can we maybe add an example of that?

I'll add it in a new PR if you don't mind, so we can merge the structure and later i'll make everything work

adamgoth:
> > Assuming we'll want to make use of packages like this, I think this looks like a good setup. Nice work.
> > I noticed the README talks about using the other packages, such as hooks, but it doesn't look like the PR goes as far as importing the package anywhere in the app (from what I can tell).
> > Can we maybe add an example of that?
> 
> I'll add it in a new PR if you don't mind, so we can merge the structure and later i'll make everything work

In that case, the structure looks good to me. But let's let Phil/Tyler have a chance to review before merging.

PR #19, opened 2023-01-10
reorganize code in packages

rafinskipg:
- Creates "@makerdao-dux/contracts" and "@makerdao-dux/hooks"

The website uses both dependencies, hooks uses contracts dependencies.


apps/webapp/src/modules/tokens/constants/tokens.constants.ts
```diff
@@ -13,54 +13,54 @@ type TokenMapping = {
 export const TOKENS: TokenMapping = {
   [chainId.mainnet]: {
     dai: {
-      ...contracts[chainId.mainnet].DAI,
+      ...(contracts[chainId.mainnet] as any).DAI,
```
adamgoth, line 16: Do we not have a contract type yet?

apps/webapp/src/pages/Oracles.tsx
```diff
@@ -1,43 +1,11 @@
...
-import { formatEther, hexDataSlice } from 'ethers/lib/utils.js';
-import { BigNumber } from 'ethers';
-import { ORACLE_CONTRACTS } from '../modules/oracles/constants/oracles.constants';
-import { contracts } from '../modules/contracts';
-import useSWR from 'swr';
+
+import { useOracles } from '@makerdao-dux/hooks';
 
 function Oracles(): React.ReactElement {
   // TODO move to hooks package
```
adamgoth, line 8: ✂️

packages/contracts/package.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "@makerdao-dux/contracts",
+  "version": "1.0.0",
```
adamgoth, line 3: Not super important right now but should we maybe version `0.1.0` or something for now?

adamgoth (approved):

packages/contracts/package.json
rafinskipg: yeah, we should have 0.0.1

apps/webapp/src/modules/tokens/constants/tokens.constants.ts
rafinskipg: I will fix this, is because i doubt we will have this tokens module here, i lean toward removing it and use the other packages better

PR #23, opened 2023-01-12
Feat/add typecheck

b-pmcg:
### What does this PR do?
- Adds typecheck script to packages and apps
- moves tsconfig.json from webapp src, to webapp root
- fixes some type errors


rafinskipg (approved):
LGTM

PR #30, opened 2023-01-17
Dsr hooks interface

rafinskipg:
Refactor DSR and DSProxy hooks interfaces using the "isLoading, error, mutate, data" definitions.

Attempt to type useContractRead responses by using typed/const ABIS, still not definitive approach, but mergeable.


packages/hooks/src/dsr/useDsr.ts
```diff
@@ -1,51 +1,122 @@
...
   });
 
+  // Hook common interface
+  const isLoading = useMemo(() => {
+    return chiIsLoading || PieIsLoading || dsrIsLoading || pieIsLoading;
+  }, [chiIsLoading, PieIsLoading, dsrIsLoading, pieIsLoading]);
+
+  const error = useMemo(() => {
+    return chiError && pieError && PieError && dsrError;
+  }, [chiError, pieError, PieError, dsrError]);
```
adamgoth, line 85: I'm not sure we want to group all these together like this. If one thing takes a long time to load, or one thing errors for some reason, we lose access to all the other data to display on the page.

I was actually debating whether each of these should be their own hook that `useDsr` then uses.

packages/hooks/src/dsr/useDsr.ts
```diff
@@ -1,51 +1,122 @@
...
+    refetchChi();
+    refetchDsr();
+    refetchPie();
+    refetchpie();
+  };
+
+  const data = useMemo<DSRHookData | undefined>(() => {
+    if (!pie || !Pie || !dsr || !chi) {
+      return undefined;
+    }
```
adamgoth, line 97: Similar to previous comment but I think we should be displaying data when it becomes available. Some stuff may take longer to fetch/calculate than others. Wdyt?

rafinskipg: I'm unsure. Depends how the useDSR hook is going to be used, if you need all the variables to compute the result, i would say wait for all of them. But if not, and you might need some of them separately, maybe creating different hooks is the best approach

packages/hooks/src/dsr/useDsr.ts
rafinskipg: I was going to export the logic to fetch everything under one async function and catch any "error" as one single output response, but i thought it was complicating a lot and since we were already using useContractRead, just add a computed error. 

If one of these calls errors, we asume the whole hook errors since, it exports all the data. Maybe we should split the logic. Feel free to update it

packages/hooks/src/dsr/useDsr.ts
```diff
@@ -1,115 +1,123 @@
...
     functionName: 'pie',
     args: [dsProxy?.dsProxyAddress]
   });
+  // Hook common interface
+  const isLoading = useMemo(() => {
+    return chiIsLoading || PieIsLoading || dsrIsLoading || pieIsLoading;
+  }, [chiIsLoading, PieIsLoading, dsrIsLoading, pieIsLoading]);
+
+  const error = useMemo(() => {
+    return chiError && pieError && PieError && dsrError;
```
adamgoth, line 81: I think these `&&`'s should be `||`s

packages/hooks/src/dsr/useDsr.ts
```diff
@@ -1,115 +1,123 @@
...
+    data: Pie,
+    isLoading: PieIsLoading,
+    error: PieError,
+    refetch: refetchPie
+  } = useContractRead({
     address: contracts[chainId]?.MCD_POT?.address,
-    abi: contracts[chainId]?.MCD_POT?.abi,
+    abi: ABIS.MCD_POT,
     functionName: 'Pie'
   });
```
adamgoth, line 51: I thought we talked about making each of these their own hooks, in case a user wanted to access just the single data point.

adamgoth (commented):
I guess I'm still not 100% convinced that having something like `{ data, isLoading, error } = useDsr()` is better than `{ chi, tvl, pie } = useDsr()`, where `chi`, `tvl`, etc all have conform to an interface of `data`, `isLoading`, and `error`.

So 
```
{
  chi: {
    data,
    isLoading,
    error
  },
   tvl: {
    data,
    isLoading,
    error
  }
```

vs

```
data: {
 chi: BigNumber,
 tvl: BigNumber,
},
error: Boolean
isLoading: Boolean
}

My issue with the latter is that you don't know what is still loading or what has an error. And you don't get any data back until everything is loaded. It obscures the data and loading/error states more than the former.

When we have larger data in hooks like delegates with stats etc and vaults, it seems a single error would prevent all data from returning.

rafinskipg:
> now what is still loading or what has an error. And you don't get any data back until everything is loaded. It obscures the data and loading/error states more than the former.
> 
> When we have larger data in hooks like delegates w

I think hooks have to be simple and try to comply to a common interface, specially if other developers are going to use it. 

If you are using the DSR hook, you want to access to the DSR data. If you want the chi data you should use the chi hook. 

If the hooks internally have some sort of cache, like swr (or if wagmi useContractRead has it's own internal cache), there aren't goingto be too many aditional queries.

rafinskipg:
> As a side note, the DSR balance isn't loading and the buttons don't seem to trigger any transactions. I took a quick look and the data coming back from `useContractWrite` seemed to be missing but not sure why.

i'll take a look

packages/hooks/src/dsr/useDsr.ts
rafinskipg: RIGHT!

packages/hooks/src/dsr/useDsr.ts
rafinskipg: yeah i had that missing , i didn't finished it (mentioned to you before) , i'm pushing it now

PR #32, opened 2023-01-19
Dsr UI updates

adamgoth:
https://www.loom.com/share/cd14cfc32c2340718b183ccb56e1cd14


apps/webapp/src/modules/ui/components/SimpleModal.tsx
```diff
@@ -0,0 +1,54 @@
+import {
```
rafinskipg, line 1: is there a complex modal? 

if not, i would suggest to just name it Modal

apps/webapp/src/modules/ui/components/SimpleModal.tsx
```diff
@@ -0,0 +1,54 @@
...
+  primaryButtonText: string;
+  secondaryButtonText: string;
+}) {
+  return (
+    <Modal isOpen={isOpen} onClose={onClose}>
+      <ModalOverlay />
+      <ModalContent>
+        <ModalHeader>{title}</ModalHeader>
+        <ModalCloseButton />
+        <ModalBody>{description}</ModalBody>
```
rafinskipg, line 40: What if we pass just a "children" component? 
Then we can have a "approve/reject" form component with all this logic.

rafinskipg (approved):
LGTM. i added some minor codes regarding modal naming , and child component passing. 

There's a little error on the linter i believe

PR #33, opened 2023-01-23
Adding datasources rendering and information

rafinskipg:
## Summary
Split "chainId" constants into a new package called "@makerdao-dux/utils" and added a new shared util called "getEtherscanLink" that gives you an etherscan link for any tx/network.
Added "dataSources" export to all the dsr and other hooks.
In the webapp added a new component called datasource, that can be used to wrap any UI component and display data sources when user hovers over the data.
## Changes Made
Created a new package called "@makerdao-dux/utils" to house the "chainId" constants and "getEtherscanLink" util.
Added "dataSources" export to all the dsr and other hooks.
Implemented a new component called "datasource" in the webapp to display data sources when user hovers over the data.
## Reasoning
The "getEtherscanLink" util is a useful tool for developers and should be easily accessible. Splitting the "chainId" constants into a separate package allows for easy maintenance and organization. Adding the "dataSources" export to hooks will provide useful information to users and the "datasource" component in the webapp will display this information in a user-friendly way.

## Additional Context
This pull request is based on Adam's idea from last week's messages. The package name "@makerdao-dux/utils" is open to change, and can be re-evaluated later as more components are added. The UI design for the "datasource" component is also open to change and can be discussed further.


apps/webapp/src/modules/ui/components/DataSourceInfo.tsx
```diff
@@ -0,0 +1,116 @@
...
+          alignItems: 'center',
+          justifyContent: 'flex-end',
+          paddingRight: '10px',
+          cursor: 'pointer',
+          position: 'absolute',
+          bottom: '0',
+          right: '0'
+        }}
+        onClick={onOpen}
+      >
```
adamgoth, line 66: These `div`'s should be using style system components. Adding styles manually like this is what we want to avoid.

PR #34, opened 2023-01-24
Subgraph delegates hook call

rafinskipg:
Added a hook called "useDelegates" that retrieves paginated data for delegates. The hook can be invoked some parameters for page and pageSize (followed advanced practices case from SWR: https://swr.vercel.app/docs/pagination) . Additionally, a lightweight client (https://www.npmjs.com/package/graphql-request) was chosen for querying the subgraph, and a "load more" button was added to the UI for displaying the list of delegates.

Messari subgraph: https://thegraph.com/hosted-service/subgraph/messari/maker-governance


tyler17 (approved):
just a couple questions, but otherwise looks good:
-I believe there's only 167 delegate contracts, but I'm able to load at least a few hundred in the app - why is that?
-I previously thought you said that a limitation of the subgraph is that it wouldn't return delegates with 0 MKR delegation history. But it seems that's actually not an issue? That's great.

rafinskipg:
> tracts, but I'm able to load at least a few hundred in the app - why is that?
> -I previously thought you said that a limitation of the subgraph is that it wou

hey @tyler17 i believe the limitation is still there. the thing is , if a delegate received MKR at any point, it gets created, then if the MKR is removed, is still there (the entity). 

But if a delegate never receives MKR , it does not get create, since it listens to that event

rafinskipg:
Added a new PR to this PR that uses mock data : https://github.com/makerdao-dux/metadao-frontend/pull/35

apps/webapp/package.json
```diff
@@ -24,8 +24,8 @@
     "@emotion/react": "^11",
     "@emotion/styled": "^11",
     "@makerdao-dux/contracts": "workspace:^",
-    "@makerdao-dux/utils": "workspace:^",
     "@makerdao-dux/hooks": "workspace:^",
+    "@makerdao-dux/utils": "workspace:^",
```
adamgoth, line 28: Let's not forget to add this to the README with the other packages

adamgoth (approved):
LGTM, nice work

PR #35, opened 2023-01-26
Use mock delegate data

rafinskipg:
Reads a mock delegate data (from static files since the arweave endpoint does not support CORS), and returns the front-matter parsed to be represented on the UI


apps/webapp/src/modules/delegates/components/DelegatesList/DelegateListItem.tsx
```diff
@@ -9,6 +9,13 @@ export function DelegateListItem({ delegate }: { delegate: Delegate }): React.Re
         <Text>Num votes: {delegate.numberVotes}</Text>
         <Text>MKR Delegated: {delegate.votingPower}</Text>
       </CardBody>
+      {delegate.delegateInfo && (
+        <CardFooter>
+          <Text>Name :{delegate.delegateInfo?.name}</Text>
+          <Text>Description: {delegate.delegateInfo?.description}</Text>
+          <img src={delegate.delegateInfo?.image} />
+        </CardFooter>
+      )}
```
adamgoth, line 18: Maybe better to add a loading state here if no `delegateInfo` yet, so we don't get cards changing size?

packages/hooks/src/decentralizedStorage/useDecentralizedStorage.ts
```diff
@@ -0,0 +1,47 @@
...
+type DecentralizedStorageHookResponse = MakerReadHook & {
+  data?: string | Blob;
+};
+
+function fetchFile(hash: any) {
+  return Promise.resolve(hash);
+}
+
+// - fileHash: the hash of the file to fetch
+export function useDecentralizedStorage(fileHash: string): DecentralizedStorageHookResponse {
```
adamgoth, line 31: Maybe better to create specific ones for different storage options, unless it will support multiple.

`useIpfsStorage` or `useArweaveStorage` for example

apps/webapp/src/modules/delegates/components/DelegatesList/DelegateListItem.tsx
rafinskipg: If there's no delegateInfo it should be a shadow delegate right now. And in theory the delegate and delegateInfo should come at the same time. 

But if no delegate is present, we should put some loading state yes. I tried adding the skeleton but nothing shows. I don't know if it's because is the same color than the page or what is happening

packages/hooks/src/decentralizedStorage/useDecentralizedStorage.ts
rafinskipg: Agree! 100% will change this once we start implementing any ipfs or arweave client. I want to wait to see what @tyler17  mentions about using or how to use arweave

PR #36, opened 2023-01-26
Eslint and e2e updates

adamgoth:
Minor README updates

Updates to fix linting errors


apps/webapp/.eslintrc.json
```diff
@@ -36,6 +36,7 @@
     "@typescript-eslint/ban-ts-comment": "warn",
     "@typescript-eslint/no-explicit-any": "off",
     "@typescript-eslint/no-unused-vars": "error",
+    "@typescript-eslint/no-var-requires": "off",
```
rafinskipg, line 39: i think we might need to move the eslint to the root project, and if the webapp has different eslint rules, extend it,

apps/webapp/package.json
```diff
@@ -60,12 +60,15 @@
...
     "@testing-library/react": "^13.3.0",
     "@types/cypress": "^1.1.3",
+    "@types/jest": "^29.4.0",
     "@types/react-helmet": "^6.1.6",
     "@typescript-eslint/eslint-plugin": "^5.48.0",
     "@typescript-eslint/parser": "^5.48.0",
     "@vitejs/plugin-react": "^2.2.0",
     "@vitest/coverage-c8": "^0.27.0",
     "babel-loader": "^8.3.0",
+    "cypress": "^12.4.0",
```
rafinskipg, line 70: do we need cypress even we have synpress installed that is using cypress also?

I see synpress uses cypress 11 so installing the 12 also will add an additional dev dependency. Is it because typescript doesnt pick well some types?

rafinskipg (changes requested):

apps/webapp/package.json
adamgoth: Yeah, couldn’t figure out how to get the types without doing that

rafinskipg (approved):

rafinskipg (changes requested):

apps/webapp/tests/e2e/specs/dsr.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+
+    cy.contains('MetaMask').should('be.visible');
+    cy.wait(1000).then(() => {
+      cy.contains('MetaMask').click();
+
+      cy.acceptMetamaskAccess().then(connected => {
+        expect(connected).to.be.true;
+      });
+    });
+    cy.get('h2').contains('Manage').should('be.visible');
```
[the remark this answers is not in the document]
adamgoth: Good catch

packages/hooks/src/dsr/useDsrWithdraw.ts
```diff
@@ -20,7 +20,9 @@ export function useDsrWithdraw({
   amount: string;
 }): MakerWriteHook {
   const network = useNetwork();
-  const chainId = network.chain?.id || 1;
+  const chain = network.chain?.id || 1;
+  // if we are on hardhat, use mainnet contracts
+  const chainId = chain === 31337 || chain === 1337 ? 1 : chain;
```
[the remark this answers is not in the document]
adamgoth: Good catch, this was meant to be a temporary workaround

PR #39, opened 2023-01-31
Add ds proxy flow, add DSR flow e2e test

adamgoth:
### What does this PR do?

Adds the code needed for creating a DS proxy wallet

Adds it to the DSR deposit UI flow

Adds an e2e test going through this flow


rafinskipg (approved):

apps/webapp/src/modules/dsr/components/DsrDepositModal.tsx
```diff
@@ -5,29 +5,45 @@ import {
...
 export function DsrDepositModal() {
   const { isOpen, onOpen, onClose } = useDisclosure();
   const [depositAmount, setDepositAmount] = React.useState<string>('');
   const toast = useToast();
-  const { mutate: mutateDsr } = useDsrData();
+  const { mutate: mutateDsr, data: dsrData } = useDsrData();
   const { data: allowance, mutate: mutateAllowance } = useDsrAllowance();
   const [hasAllowance, setHasAllowance] = React.useState<boolean>(false);
+  const dsProxy = useDsProxyData();
+  const hasDsProxy = dsProxy?.data !== ZERO_ADDRESS && dsProxy?.data;
```
tyler17, line 34: It seems not ideal to expect the user of the hooks library to know to null check with the ZERO_ADDRESS here?

apps/webapp/src/pages/Account.tsx
```diff
@@ -0,0 +1,60 @@
...
+        title: 'DS proxy creation failed',
+        description: 'Failed to create a DS proxy wallet',
+        status: 'error',
+        duration: 5000,
+        isClosable: true,
+        position: 'top-right'
+      });
+    }
+  });
+  const dsProxyAddress = dsProxy?.data !== ZERO_ADDRESS && dsProxy?.data;
```
tyler17, line 35: same comment here about the zero address

tyler17 (approved):
just had one comment about the zero address, otherwise looks good

apps/webapp/src/modules/dsr/components/DsrDepositModal.tsx
adamgoth: Agree, I’ll move this logic into the hook

PR #41, opened 2023-02-02
Use ipfs real data

rafinskipg:
IPFS fetch real file with all the delegates and cross merge info with the messari delegates

image


apps/webapp/src/modules/utils/markdownToHTML.ts
```diff
@@ -0,0 +1,19 @@
+import remarkGfm from 'remark-gfm';
+
```
rafinskipg, line 2: Ported from gov-portal

apps/webapp/src/modules/delegates/components/DelegateIconAndName.tsx
```diff
@@ -0,0 +1,28 @@
...
+export function DelegateIconAndName({ delegate, ...props }: { delegate: Delegate }): React.ReactElement {
+  return (
+    <Flex {...props} alignItems="center">
+      {delegate.delegateMetadata?.image ? (
+        <Image
+          src={`https://makerdao-dux.infura-ipfs.io/ipfs/${delegate.delegateMetadata?.image}`}
+          alt="delegate-icon"
+          boxSize="40px"
+          borderRadius="full"
+          mr="10px"
```
adamgoth, line 13: Small thing but we should try to stick to spacing units as part of the style system. Makes it more consistent and makes it easier to update app-wide if we need to compared to using an arbitrary number of `px`.

https://chakra-ui.com/docs/styled-system/theme#spacing

apps/webapp/src/modules/delegates/components/DelegateIconAndName.tsx
```diff
@@ -0,0 +1,28 @@
...
+        <Image
+          src="https://via.placeholder.com/150"
+          alt="delegate-icon"
+          boxSize="40px"
+          borderRadius="full"
+          mr="10px"
+        />
+      )}
+
+      <Text fontWeight={'bold'}>{delegate.delegateMetadata?.profile.name || 'Shadow Delegate'}</Text>
```
adamgoth, line 25: We should try to stick to a predefined text variant here rather than adding custom styles/font weights.

apps/webapp/src/modules/delegates/components/DelegatesList/DelegateListItem.tsx
```diff
@@ -1,19 +1,45 @@
-import { Card, CardBody, CardFooter, Text } from '@chakra-ui/react';
+import { Card, CardBody, CardFooter, CardHeader, Text, Box, Flex } from '@chakra-ui/react';
 import { Delegate } from '@makerdao-dux/hooks';
-
+import { Link } from '@chakra-ui/react';
```
adamgoth, line 3: This can go in the other `'@chakra-ui/react'` import on line 1.

apps/webapp/src/modules/delegates/components/DelegatesList/DelegateListItem.tsx
```diff
@@ -1,19 +1,45 @@
...
+import { useEffect, useState } from 'react';
+import { markdownToHtml } from '../../../utils/markdownToHTML';
 export function DelegateListItem({ delegate }: { delegate: Delegate }): React.ReactElement {
+  const [htmlDescription, setHTMLDescription] = useState('');
+
+  useEffect(() => {
+    const parseHTMLDescription = async () => {
+      const description = await markdownToHtml(delegate.delegateMetadata?.profile.description || '');
+      setHTMLDescription(description);
+    };
```
adamgoth, line 18: Should we move this out of the useEffect so that it doesn't get redefined each time? Can pass the `delegate` as an argument.

apps/webapp/src/modules/delegates/components/DelegatesList/DelegateListItem.tsx
```diff
@@ -1,19 +1,45 @@
...
   return (
     <Card data-testid="delegate-card">
+      <CardHeader>
+        <Box>
+          <Text>MKR: {formatUnits(delegate.votingPowerRaw)}</Text>
+        </Box>
+        <Flex>
+          <Box>
+            <Link href={getEtherscanLink(chainId.mainnet, delegate.address, 'address')} isExternal>
+              Delegate Contract: {formatAddress(delegate.address)} <ExternalLinkIcon mx="2px" />
```
adamgoth, line 30: Try a spacing unit instead of px

packages/hooks/src/decentralizedStorage/useIPFSStorage.ts
```diff
@@ -23,12 +23,16 @@ type DecentralizedStorageHookResponse = MakerReadHook & {
   data?: string | Blob;
 };
 
-function fetchFile(hash: any) {
-  return Promise.resolve(hash);
+function fetchFile(hash: string) {
+  const ipfsGateway = `https://makerdao-dux.infura-ipfs.io/ipfs/${hash}`;
```
adamgoth, line 27: I guess we can update this later when we have a better idea of what gateways we'll be using, but wondering if the gateway should be a parameter to `fetchFile` and we can have a set of predefined ones.

adamgoth: Should the `href` on the `dataSource` info we return match this?

adamgoth (commented):
This is cool to see. Seems to load well, but recognized and shadow delegates seem mixed together. I guess doesn't matter for a PoC though. I left some minor comments/questions.

apps/webapp/src/modules/delegates/components/DelegateIconAndName.tsx
adamgoth: This probably should be a `<Heading />` I guess.

apps/webapp/src/modules/delegates/components/DelegateIconAndName.tsx
rafinskipg: oki will change it tomorrow

apps/webapp/src/modules/delegates/components/DelegateIconAndName.tsx
rafinskipg: oki

apps/webapp/src/modules/delegates/components/DelegatesList/DelegateListItem.tsx
rafinskipg: i'm not sure which one is the most efficient way of doing it, maybe defining outside of the component, i'll change it tomorrow

packages/hooks/src/delegates/useDelegatesMetadata.ts
```diff
@@ -0,0 +1,46 @@
...
+import { MakerReadHook } from '../hooks';
+import { DelegateMetadata, DelegateTag } from './delegates';
+
+type UseDelegatesMetadataResponse = MakerReadHook & {
+  data?: DelegateMetadata[];
+};
+
+// Returns the delegates metadata fetched from IPFS
+export function useDelegatesMetadata(): UseDelegatesMetadataResponse {
+  const response = useIPFSStorage('QmVnBkMtiP13HYw4HueBV8sgMUXMC7RKnDQXhXBo22n9se');
```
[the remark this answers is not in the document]
tyler17: will push up a PR that reads this hash from ens shortly

tyler17 (approved):
Really nice work.  Having everything in one IPFS file seems to simplify things.  When I view the preview I don't see the metadata information for GFX Labs or Flip Flap Flop, but I do for the other delegates - is that expected?

packages/hooks/src/decentralizedStorage/useIPFSStorage.ts
rafinskipg: Merged both datasources

rafinskipg:
> Really nice work. Having everything in one IPFS file seems to simplify things. When I view the preview I don't see the metadata information for GFX Labs or Flip Flap Flop, but I do for the other delegates - is that expected?

I guess there's some problem with what the graph is returning and the hash that we get from the community repo. 

We didn't account for renewed delegate contracts, so we have the old address. We had that mapping on the gov portal

PR #42, opened 2023-02-02
Use ens

rafinskipg:
@tyler17 useENs


rafinskipg (commented):
Did you check if any of the useENS* hooks from wagmi work? 

https://wagmi.sh/docs/migration-guide#useenslookup

packages/hooks/src/decentralizedStorage/useEns.ts
```diff
@@ -0,0 +1,51 @@
...
+    data: ipfsHash,
+    error: ensError,
+    mutate: ensMutate,
+    isValidating: ensIsValidating
+  } = useSWR('ens', () => resolver?.getText('ipfs'));
+
+  const mutate = () => {
+    resolverRefetch();
+    ensMutate();
+  };
```
rafinskipg, line 28: I think you should use an useMemo instead of useSWR, because calling the 2 mutates at the same time won't execute them sequentially. 

whenever the resolver changes, the ipfs text will be updated


```
const txt = useMemo(() => {

return resolver?.getText('ipfs')
}, [resolver])


```

packages/hooks/src/delegates/useDelegatesMetadata.ts
```diff
@@ -2,15 +2,17 @@ import { useMemo } from 'react';
...
 
 type UseDelegatesMetadataResponse = MakerReadHook & {
   data?: DelegateMetadata[];
 };
 
 // Returns the delegates metadata fetched from IPFS
 export function useDelegatesMetadata(): UseDelegatesMetadataResponse {
-  const response = useIPFSStorage('QmVnBkMtiP13HYw4HueBV8sgMUXMC7RKnDQXhXBo22n9se');
-
+  const { data: ensHash, dataSources: ensDataSources } = useEns();
```
rafinskipg, line 13: maybe call it useENSIPFSHash or something, and pass the ENS name as a parameter to the hook

packages/hooks/src/decentralizedStorage/useEns.ts
tyler17: Hm well I'm not sure how to make this work with a useMemo though, because using the above code, txt will be a promise not the actual text  @rafinskipg

packages/hooks/src/decentralizedStorage/useEns.ts
```diff
@@ -0,0 +1,51 @@
...
+    isLoading: resolverIsLoading,
+    error: resolverError,
+    refetch: resolverRefetch
+  } = useEnsResolver({ name, chainId });
+  const {
+    data: ipfsHash,
+    error: ensError,
+    mutate: ensMutate,
+    isValidating: ensIsValidating
+  } = useSWR('ens', () => resolver?.getText('ipfs'));
```
adamgoth, line 23: I'm confused how `getText('ipfs')` works. Is `ipfs` the key of the text record? I thought it was `url`.

packages/hooks/src/delegates/useDelegatesMetadata.ts
adamgoth: Or if we want it to be more reuseable, maybe `useEnsTextRecord`, and pass both ENS name and name of the text record?

packages/hooks/src/delegates/useDelegatesTags.ts
```diff
@@ -2,15 +2,17 @@ import { useMemo } from 'react';
...
 type UseDelegateTagsResponse = MakerReadHook & {
   data?: DelegateTag[];
 };
 
 // Returns the delegates metadata fetched from IPFS
 export function useDelegatesTags(): UseDelegateTagsResponse {
-  const response = useIPFSStorage('QmVnBkMtiP13HYw4HueBV8sgMUXMC7RKnDQXhXBo22n9se');
-
+  const { data: ensHash, dataSources: ensDataSources } = useEns();
+  const response = useIPFSStorage(ensHash || ''); //TODO add a backup hash stored in local filesystem?
```
adamgoth, line 14: Hmm, I think we should probably update the IPFS hook and all other downstream hooks to be able to handle a missing hash. I'm not sure if it's a good idea to display a backup hash, because it may contain outdated or wrong information which could be confusing if that gets displayed. Wdyt?

packages/hooks/src/decentralizedStorage/useEns.ts
rafinskipg: alright then in that case it should be a useEffect that does an async call , memo won't work with async

packages/hooks/src/delegates/useDelegatesMetadata.ts
rafinskipg: yes better

packages/hooks/src/delegates/useDelegatesTags.ts
rafinskipg: i think we shouldn't have a backup hash on any hook, just return an error if the data is not fetched well

packages/hooks/src/decentralizedStorage/useEns.ts
tyler17: Yeah I set ipfs as the key of the text record

packages/hooks/src/delegates/useDelegatesTags.ts
tyler17: Makes sense.  I was just trying to speed up the time it takes to show some metadata, but maybe we could just start the ens fetch as soon as you visit the app, instead of waiting to for the user to visit the delegates page.

tyler17: Well actually it loads the ens text record pretty quickly for me now, not sure why it was slow for me earlier, maybe just getting rate limited by the default provider

tyler17: Regarding making downstream hooks handle a missing hash, it doesn't seem to be a problem because we're using the ipfs hash as the useSwr key, and swr doesn't fetch the data if the key is falsey (https://github.com/vercel/swr/discussions/1810#discussioncomment-2049694) I added a comment about this in the code, but let me know if you think there's something that should be done

packages/hooks/src/decentralizedStorage/useEns.ts
tyler17: done 👍

packages/hooks/src/delegates/useDelegatesMetadata.ts
tyler17: Good idea, updated it

adamgoth (approved):

tyler17:
Thanks for the Context/Provider update @rafinskipg.  I see the typecheck is failing with `Module '"@makerdao-dux/hooks"' has no exported member 'MakerHooksProvider'.` but not sure why since it does appear to be exported in the index file.

rafinskipg:
> Thanks for the Context/Provider update @rafinskipg. I see the typecheck is failing with `Module '"@makerdao-dux/hooks"' has no exported member 'MakerHooksProvider'.` but not sure why since it does appear to be exported in the index file.

is because the NPM cache from github, clearing cache https://github.com/makerdao-dux/metadao-frontend/actions/caches here , fixes it

tyler17:
> > Thanks for the Context/Provider update @rafinskipg. I see the typecheck is failing with `Module '"@makerdao-dux/hooks"' has no exported member 'MakerHooksProvider'.` but not sure why since it does appear to be exported in the index file.
> 
> is because the NPM cache from github, clearing cache https://github.com/makerdao-dux/metadao-frontend/actions/caches here , fixes it

Ah nice!

PR #47, opened 2023-02-07
Add tokeninput, fix storybook

rafinskipg:
Adds a token input that accepts decimals with commas or dots. Gives errors with negative numbers and returns always a bignumber.

This component is meant to be used with the UI for inputting tokens


apps/webapp/src/modules/ui/components/TokenInput.tsx
```diff
@@ -0,0 +1,73 @@
...
+        setInputValue(formatUnits(value));
+      }
+    } catch (e) {
+      setErrorInvalidFormat(true);
+      onChange(parseUnits('0'));
+    }
+  }, [value]);
+
+  return (
+    <React.Fragment>
```
adamgoth, line 60: `<React.Fragment>` seems unneeded here. 

I think we might be able to wrap this in chakra's `FormControl` component and pass `errorInvalidFormat` to it. Then it will inherit the error styling and be consistent with other form elements.

apps/webapp/src/modules/ui/components/TokenInput.tsx
```diff
@@ -0,0 +1,73 @@
...
+      <Input
+        data-testid="token-input"
+        sx={{
+          borderColor: errorInvalidFormat ? 'red' : 'default'
+        }}
+        type="text"
```
adamgoth, line 66: Shouldn't this be `type="number"` for a token input?

Testing via storybook, it seems you can type anything in.

Screenshot 2023-02-07 at 11 46 25 AM
Screenshot 2023-02-07 at 11 47 21 AM

adamgoth (changes requested):
Can we hook it up to DSR so we can make sure everything works end-to-end and consider the bug resolved?

Happy to jump in and help get it over the line, but let me know your thoughts on other PR comments.

apps/webapp/src/modules/ui/components/TokenInput.tsx
rafinskipg: We can do type number, but then it won't allow to add comas or dots at the same time, only numbers (which i think is fine too), first implementation i did allowed only for numbers

rafinskipg:
> Can we hook it up to DSR so we can make sure everything works end-to-end and consider the bug resolved?
> 
> Happy to jump in and help get it over the line, but let me know your thoughts on other PR comments.

Oki, I'll add it tomorrow morning

adamgoth (approved):
LGTM, this feels much better using BigNumber

I just swapped out the helper text for error text on the error message and added `formatUnits` to a few more places in the modals that needed it

PR #49, opened 2023-02-08
Add makerdao hooks provider

rafinskipg:
Introduces `<MakerDaoHooksProvider config={}>` provider that has to be used to wrap the app. It's the same concept like "WagmiConfig".

This helps reduce "constants" that shouldn't be tied to the hooks package, like the ENS name or the IPFS gateway.

It works well. But we should in the future change the structure of the delegates object to add ENS for each chain. But it works for now.


tyler17:
What do you mean by "ens for each chain"?

PR #52, opened 2023-02-13
Allow users to set up IPFS gateway

rafinskipg:
Allows users to set up the IPFS gateway on the config page

image


tyler17 (approved):
Looks good 👍  In the future we could add a test button like with the rpc

PR #53, opened 2023-02-13
Update RPC settings logic, add RPC test button

adamgoth:
### What does this PR do?

Updates how we handle settings changes in the form

Adds a button to test RPC connection (no UI added for response, just console log for now)


apps/webapp/src/modules/config/components/RPCSetting.tsx
```diff
@@ -14,10 +14,34 @@ export function RPCSetting({
...
+      handleRpcChange({ chainId, url: defaultRPC });
+    }
+    setOption(option == RpcStatus.Custom ? RpcStatus.Public : RpcStatus.Custom);
+  };
+
+  const testRpcConnection = async (url: string) => {
+    try {
+      const res = await fetch(url, {
+        method: 'POST',
+        body: JSON.stringify({ method: 'eth_blockNumber', params: [], id: 42, jsonrpc: '2.0' })
```
tyler17, line 30: 42 here means we're testing with kovan?

tyler17 (approved):

apps/webapp/src/modules/config/components/RPCSetting.tsx
rafinskipg: mhh yeah, wondering the same

adamgoth: Good question. It's actually just used as an identifier to match request and response. I had copied it from a request to mainnet in the network tab. The chain for the network is typically in the URL.

Screenshot 2023-02-14 at 7 57 09 AM

Screenshot 2023-02-14 at 7 57 51 AM

PR #54, opened 2023-02-14
Trying to set up accounts on the tests

rafinskipg:
It tries to add tests to the hooks package. The goal is to be able to mock contract calls or test things correctly. 

I believe wagmi uses a local test network when testing the hooks. We should think about it. 

We want to be able to test that we can deposit/ withdraw , create proxies, etc etc


tyler17:
Nice! Maybe we should add an alchemy / infura key that can be used in the tests?

tyler17 (approved):

PR #58, opened 2023-02-20
Contenthash

tyler17:
use ens contenthash instead of ens text record


packages/hooks/src/decentralizedStorage/UseEnsContent.ts
```diff
@@ -4,32 +4,37 @@ import { useState, useEffect } from 'react';
...
+}
+
+type UseEnsContentProperties = {
   name: string;
-  key: ensTextRecordKey;
+  subdomain: ensSubdomain;
   chainId?: number;
 };
-export function useEnsTextRecord({
+export function UseEnsContent({
```
rafinskipg, line 19: it should be `useEnsContent` (camelCase)

tyler17: good catch, thanks

rafinskipg (approved):

adamgoth (approved):

PR #59, opened 2023-02-20
create delegate detail route and fix some dark mode issues

rafinskipg:
### What does this PR do?

Adds a new route : 'delegates/address'
Adds the useDelegate(address) hook to fetch only one delegate
Fixes some dark mode issues

image


apps/webapp/src/pages/DelegateDetail.tsx
```diff
@@ -0,0 +1,40 @@
...
+  const [htmlDescription, setHTMLDescription] = useState('');
+
+  useEffect(() => {
+    const parseHTMLDescription = async () => {
+      if (!data) return;
+      const description = await markdownToHtml(data.delegateMetadata?.profile.description || '');
+      setHTMLDescription(description);
+    };
+    parseHTMLDescription();
+  }, [data]);
```
torhector2, line 24: This could be a hook on its own. Looks like we could use it on other components.
Something like `useMakdownToHtml`

apps/webapp/src/pages/main.tsx
```diff
@@ -3,8 +3,8 @@ import ReactDOM from 'react-dom/client';
 import { theme } from '../modules/ui/theme';
 import { WagmiConfig } from 'wagmi';
 import { getWagmiClientAndChains } from '../modules/providers/wagmi';
-import { lightTheme, RainbowKitProvider, Theme } from '@rainbow-me/rainbowkit';
-import { ChakraProvider, ColorModeScript /*, useColorMode */ } from '@chakra-ui/react';
+import { darkTheme, lightTheme, RainbowKitProvider, Theme } from '@rainbow-me/rainbowkit';
+import { ChakraProvider, ColorModeScript /*, useColorMode */, useColorMode } from '@chakra-ui/react';
```
torhector2, line 7: We can remove the commented code here

torhector2 (approved):
It looks good. I made some suggestions but they're not blockers.

tyler17 (approved):
Nice!

apps/webapp/src/pages/DelegateDetail.tsx
rafinskipg: Yeah will do!

PR #60, opened 2023-02-21
Adding docs for hooks

rafinskipg:
Adds markdown documentation for existing hooks


docs/hooks/dsr/useDsrAllowance.md
```diff
@@ -0,0 +1,33 @@
...
+
+```typescript
+import { useDsrAllowance } from './useDsrAllowance';
+
+function MyComponent() {
+  const { data: dsrAllowance } = useDsrAllowance();
+
+  if (!dsrAllowance) {
+    return <div>Loading...</div>;
+  }
```
adamgoth, line 29: Seems this example should be using the `isLoading` property for render `<div>Loading...</div>`

docs/hooks/dsr/useDsrAllowance.md
```diff
@@ -0,0 +1,33 @@
...
+
+  return <div>DSR allowance: {dsrAllowance.toString()}</div>;
```
adamgoth, line 31: Probably wrap in an `if` since `dsrAllowance` can be null

adamgoth (commented):
I spot-checked a few and think we can improve the examples slightly. Overall looks great though. Really nice to have docs for all these.

docs/hooks/delegates/useDelegate.md
```diff
@@ -0,0 +1,43 @@
...
+- `address` (`string`): The address of the delegate to fetch.
+
+### Returns
+
+An object containing the following properties:
+
+- `isLoading` (`boolean`): Indicates if the data is being loaded.
+- `error` (`Error | undefined`): If an error occurred during data fetching, it will be set here.
+- `mutate` (`(data?: Delegate, shouldRevalidate?: boolean) => Promise<any>`): A function to mutate the data.
+- `data` (`Delegate | undefined`): The delegate data.
```
b-pmcg, line 16: It would be cool if we could include, or link to, a definition of the `Delegate` data type too. Idk if chatgpt can do that or not.

tyler17 (approved):

PR #61, opened 2023-02-22
dvc

rafinskipg:
Adds boilerplate for DVC strategies fetching. Initial types and some random components not tied to design
image


tyler17:
I see lint typecheck is failing, but otherwise looks like a good start

tyler17 (approved):

rafinskipg:
> I see lint typecheck is failing, but otherwise looks like a good start


i think is a cache thing , it works locally! I think the first commit i did didn't include the hook

PR #62, opened 2023-02-22
use chakra-ui icon component for icons

b-pmcg:
### What does this PR do?
Refactors the custom Icon component to use the chakra-ui component
- allows passing in any third party icon component (including chakra's) into `as` prop:
`<Icon as={HamburgerIcon} />`
- continues to allow using our custom icons from `icons.tsx` file
`<Icon name="discord" />`
- returns a default component if icon is not found
- can be styled using chakra style props
`<Icon name="discord" boxSize={3} />`

**note**: previously we used an arbitrary property, `size`, for icon size, but now use the standard `boxSize` prop.

### Steps for testing:
Check the header or delegate page to view current icons


adamgoth (approved):
💯

PR #63, opened 2023-02-23
Fetch dvcs from ipfs

rafinskipg:
Uses latest https://github.com/makerdao-dux/community/pull/252 file


packages/hooks/src/delegateVotingCommittees/useDelegateVotingCommittees.ts
```diff
@@ -1,41 +1,38 @@
...
-
 type DelegateVotingCommitteesHookResponse = MakerReadHook & {
   data?: DelelegateVotingCommittee[];
 };
 
 export function useDelegateVotingCommittees(): DelegateVotingCommitteesHookResponse {
-  const { data, error, mutate, isValidating } = useSWR<DelelegateVotingCommittee[]>('dvcs', () =>
-    fetchDVCs()
+  const { data, error, mutate, isLoading, dataSources } = useIPFSStorage(
+    'QmQ9hBWwK4CBgXfjDTQFmku3Kwd7Dg2AVGnCbGn6diw2wi'
```
tyler17, line 12: we can use the ens hook for this now that it's updated

tyler17: I'll make a PR

packages/hooks/src/delegateVotingCommittees/useDelegateVotingCommittees.ts
```diff
@@ -1,41 +1,38 @@
...
-    isLoading: !data && isValidating,
+    isLoading,
     error,
     mutate,
-    data,
+    data: dvcs,
     dataSources: [
+      ...dataSources,
       {
         title: 'Community Github',
```
tyler17, line 37: maybe the Community Gitub data source should be moved to the useEnsContent hook?

tyler17: Or actually I guess that doesn't make sense since there could be different content on ens

packages/hooks/src/delegateVotingCommittees/useDelegateVotingCommittees.ts
tyler17: PR: https://github.com/makerdao-dux/metadao-frontend/pull/64

PR #67, opened 2023-03-01
Add types to networks

torhector2:
Adds type to network objects.

Unit tests passes but I couldn't check the e2e tests as I still need to configure them properly.


packages/contracts/src/networks/arbitrum.ts
```diff
@@ -1,6 +1,7 @@
+import { NetworkContractData } from '..';
```
rafinskipg, line 1: maybe add the path? ../index

rafinskipg (approved):
minor comment

adamgoth (approved):

packages/contracts/src/networks/arbitrum.ts
torhector2: right, the VSCode quick fix imported like this. Changing it now.

PR #70, opened 2023-03-03
UI Search bar

torhector2:
Implement search functionality in the app's search bar, allowing users to find delegates by name or address, and committees and strategies by name. Results are limited to three items with an option to load more.

https://user-images.githubusercontent.com/440694/222698147-3319d123-0865-4bfd-ac60-d04666fe9c3d.mp4


torhector2:
Maybe I can try to parallelize the  loops in the search function with a Promise.all 

```
const [delegateMatches, dvcMatches] = await Promise.all([
    searchDelegates(query, delegates),
    searchDvcs(query, dvcs)
  ]);
```
I'll try that on Monday.

b-pmcg:
nice, looks pretty slick. Only thing I noticed is that the delegate markdown is being displayed as raw markdown. I wonder if we can parse the MD with like [gray-matter](https://github.com/jonschlinkert/gray-matter) or something similar. Or maybe we could display the tags in that place? Just brainstorming, I haven't seen the designs for the search results yet.

apps/webapp/src/modules/layout/components/Header.tsx
```diff
@@ -41,12 +38,7 @@ export function Header(): React.ReactElement {
           </Box>
         </Link>
         <Box px={4} width="500px">
-          <InputGroup>
-            <InputLeftElement>
-              <Icon as={SearchIcon} />
-            </InputLeftElement>
-            <Input variant="pill" placeholder="Search delegates, strategies, PVC's, farms, vaults, etc" />
-          </InputGroup>
+          <SearchBar placeholder="Search delegates, strategies, PVC's, farms, vaults, etc" />
```
adamgoth, line 41: We can update the placeholder to match the implementation now: "Search delegates and DVCs"

apps/webapp/src/modules/utils/searchGovernanceData.ts
```diff
@@ -0,0 +1,99 @@
...
+        url: `/delegates/${address}`,
+        isExternalUrl: false,
+        extra: address
+      });
+    }
+  }
+  return matches;
+}
+
+async function searchCommitees(query: string, dvcs: DelelegateVotingCommittee[]) {
```
adamgoth, line 57: Minor thing but committees needs another `t`: `searchCommittees`

apps/webapp/src/modules/ui/components/SearchResultsList.tsx
```diff
@@ -0,0 +1,40 @@
...
+    setDisplayLimit(limit);
+  }, [results, limit]);
+
+  if (results === undefined) {
+    return null;
+  }
+
+  return (
+    <List spacing={3}>
+      {!hasResults && <ListItem>No results</ListItem>}
```
adamgoth, line 27: When there are no results found, we should append links to the list pages for delegates and DVCs. So maybe "No results" then underneath "View all delegates" and under that "View all DVCs" or something like that. We might only have the delegates page in the app for now though.

adamgoth (commented):
This looks great! 🔥

Awesome work. I left a couple small comments about minor details. 

One minor thing I noticed that is not a big deal but maybe a small UX thing that we can improve: When you search for something there is no results, then you delete the text and start typing something new, it still says "No results", but maybe we could show the "Loading" state again instead since they deleted the text that resulted in "No results"

apps/webapp/src/modules/delegates/components/DelegateAvatar.tsx
```diff
@@ -1,21 +1,6 @@
...
-          alt="delegate-icon"
-          boxSize={20}
-          borderRadius="full"
-        />
-      ) : (
-        <Image src="https://via.placeholder.com/150" alt="delegate-icon" boxSize={20} borderRadius="full" />
-      )}
-    </>
-  );
+  return <AvatarIpfs image={delegate.delegateMetadata?.image} alt="delegate-icon" boxSize={20} />;
```
rafinskipg, line 5: Why do you think it should be tied to IPFS implementation? 

I think we could just have an Avatar that accepts a https image or base64 image, any image data source. 

if tomorrow we change from IPFS to another thing, the avataripfs does not make sense anymore.

```
<Avatar src={ipfsToHttps(delegate.delegateMetadata.image} />
```

Another thing we should factor in, is we might want to use ENS images / or address icons. (this depends on product decision), but in that sense it should be something like
<Avatar address={address} img={ipfsToHTTPS(delegate.delegateMedatada.image } />

apps/webapp/src/modules/ui/components/NoResultsListItem.tsx
```diff
@@ -0,0 +1,16 @@
+import { ListItem, Text, Divider, HStack } from '@chakra-ui/react';
+import { LinkButton } from './LinkButton';
```
rafinskipg, line 2: I think this should all be located under a "search" module.

rafinskipg: waiting for other devs input on this @adamgoth  @b-pmcg @tyler17

apps/webapp/src/modules/utils/searchGovernanceData.ts
```diff
@@ -0,0 +1,99 @@
+import { Delegate, DelelegateVotingCommittee } from '@makerdao-dux/hooks';
+import { Strategy } from '@makerdao-dux/hooks/dist/delegateVotingCommittees/delegateVotingCommittee';
```
rafinskipg, line 2: move this to a "search" module, along all this implementation, is too big to be splitted into UI and utils

apps/webapp/src/modules/utils/searchGovernanceData.ts
```diff
@@ -0,0 +1,99 @@
...
+  const matches: SearchResult[] = [];
+  for (const delegate of delegates) {
+    const { delegateMetadata: { profile, image } = {}, address } = delegate;
+    const { name = '', description = '', tags } = profile || {};
+
+    if (name.toLowerCase().includes(query.toLowerCase()) || address.startsWith(query)) {
+      matches.push({
+        id: address,
+        title: name,
+        image,
```
rafinskipg, line 46: here you can transform from IPFS to https if you want

apps/webapp/src/modules/utils/searchGovernanceData.ts
```diff
@@ -0,0 +1,99 @@
...
+    }
+  }
+  return matches;
+}
+
+async function searchStrategies(query: string, strategies: Strategy[]) {
+  // TODO: How do we plan to retrieve strategies? This is a temporary solution.
+  const matches: SearchResult[] = [];
+  const uniqueStrategies = uniqBy(
+    strategies,
```
rafinskipg, line 79: strategies are always unique. 

1 dvc has 1 or more strategies, but they have to be always unique . If another DVC has a strategy with the same name or ID, is still a different strategy. 

We might need to return "Strategy name, by DVC X",

apps/webapp/src/modules/utils/uniqBy.ts
```diff
@@ -0,0 +1,11 @@
+export function uniqBy<T>(arr: T[], iteratee: string | ((item: T) => string)): T[] {
```
rafinskipg, line 1: I don't think we need it based on the strategies uniqueness

apps/webapp/src/modules/utils/useSearchGovernanceData.tsx
```diff
@@ -0,0 +1,38 @@
+import { useDelegates, useDelegateVotingCommittees } from '@makerdao-dux/hooks';
+import { useState, useEffect, useRef } from 'react';
+import { searchGovernanceData, SearchResult } from './searchGovernanceData';
+import { useDebounceQuery } from './useDebounceQuery';
+
+export function useSearchGovernanceData(query: string) {
+  const [results, setResults] = useState<SearchResult[] | undefined>();
```
rafinskipg, line 7: nice 


maybe useSearch, and type the response. Maybe also consider using useSWR to cache responses by query, so if one search repeats, it doesn't perform the logic again

rafinskipg (changes requested):

apps/webapp/src/modules/ui/components/NoResultsListItem.tsx
adamgoth: sure, probably makes sense to have a search module for all the components and utils 👍

apps/webapp/src/modules/utils/searchGovernanceData.ts
torhector2: Ok, my assumption was wrong, I thought based on the datasource I was using (the strategies coming from the request) that an strategy had only name, delegates and description (no ID is returned but I guess this will change). So I thought if 2 different DVCs have an Strategy name "Growth" it'd be confusing. 

I'll presume there are no repeated strategies and will remove the uniqBy.

Also, I like the idea of "Strategy name, by DVC X"

apps/webapp/src/modules/utils/useSearchGovernanceData.tsx
torhector2: I think useSWR caches the responses to return early the cached data but then always perform the request. So in the end it will always perform the logic.

rafinskipg: maybe, i think is fine as now since all the datasources we use are already cached by SWR, so maybe no need to add another layer

apps/webapp/src/modules/search/utils/useSearchGovernanceData.tsx
```diff
@@ -0,0 +1,38 @@
+import { useDelegates, useDelegateVotingCommittees } from '@makerdao-dux/hooks';
+import { useState, useEffect, useRef } from 'react';
+import { useDebounceQuery } from '../../utils/useDebounceQuery';
+import { searchGovernanceData, SearchResult } from './searchGovernanceData';
+
```
rafinskipg, line 5: this should go under "hooks" no utils,

apps/webapp/src/modules/search/utils/searchGovernanceData.ts
```diff
@@ -0,0 +1,90 @@
...
+async function searchStrategies(query: string, strategies: (Strategy & { dvcName: string })[]) {
+  // TODO: How do we plan to retrieve strategies? This is a temporary solution.
+  const matches: SearchResult[] = [];
+
+  for (const strategy of strategies) {
+    if (strategy.name.toLowerCase().includes(query)) {
+      matches.push({
+        id: strategy.name,
+        title: `${strategy.name} by ${strategy.dvcName}`,
+        subtitle: nodeValuesFromMarkdown(strategy.description, 'paragraph', 1)[0],
```
rafinskipg, line 83: i'm not sure what this does, does it assume certain structure on the markdown file?

Because that has not been decided, and also we don't have any way to enforce it

torhector2: Yes, in this case it returns the first N elements (in this case 1) of type paragraph in the markdown. Yes I thought the structure returned was the definitely structure, my bad. What should we do here? If it not the final decision maybe we can change it later, not sure.

apps/webapp/src/modules/search/utils/searchGovernanceData.ts
```diff
@@ -0,0 +1,90 @@
...
+    searchStrategies(
+      query,
+      dvcs.flatMap(dvc => dvc.strategies.map(strategy => ({ ...strategy, dvcName: dvc.name })))
+    )
+  ]);
+
+  return [...delegateMatches, ...dvcMatches, ...strategyMatches];
+}
+
+async function searchDelegates(query: string, delegates: Delegate[]) {
```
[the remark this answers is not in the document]
torhector2: I'll do 👍

apps/webapp/src/modules/search/utils/searchGovernanceData.ts
rafinskipg: i guess we can leave it for now but keep in mind it might not make sense in the future

apps/webapp/src/modules/search/utils/useSearchGovernanceData.tsx
torhector2: Right, I'll organize these and the other hooks and utils in the webapp  👍

rafinskipg (approved):

PR #71, opened 2023-03-09
Vault math functions

b-pmcg:
### What does this PR do?
Adds a functions to do math on vault-related data. The package uses ethers v6, but the change from ethers BigNumber to javascript BigInt is not really that impactful since they are very similar. This will future-proof us when we migrate in the main webapp (after wagmi does).

The math functions expect and return a BigInt. In the webapp (where we use ethers v5) we can convert BigNumber to BigInt easily with the built in `.toBigInt()` method, and there's no need to "re-convert" it to a BigNumber. BigInt's can go directly into the ethers `formatUnit` functions for displaying in the UI.

Also adds tests for the functions using mocked data simulating a real vault that I pulled directly from the chain.


adamgoth (approved):
This is awesome, great work 🔥  It'll be very convenient for vault code to have these formulas abstracted away into convenient functions like this.

tyler17 (approved):
Looks good 👍

PR #72, opened 2023-03-10
Search bar e2e test

torhector2:
Adds e2e test for the search bar.

It wrongly presumes that certain responses will be true in the future (like the query 'llc' returning 3 results). Maybe we can use a fixture for that particular case. I guess the case where we search '0x' is so common that it'll always return 3 or more results.

### UPDATE
I included a fixtures folder with some mock data for the search. I only use it for the 'llc' query test as I think is the only one that can break in the future.


adamgoth (approved):
LGTM 💸

tyler17 (approved):

PR #74, opened 2023-03-15
Add filtering and sorting capabilities

torhector2:
Adds a basic filter bar for delegates with:

- Same input for name and address
- Input for dvc
- A reset button

There's no design so I didn't spend much time on it yet. 
It doesn't do any filtration yet as we are moving the filters and sort data to the subgraph. Once the subgraph is implemented we will update the fetchDelegates logic.


adamgoth (commented):
LGTM so far

packages/hooks/src/delegates/useDelegates.ts
```diff
@@ -44,7 +45,14 @@ function fetchDelegates(page = 0, pageSize = 10): Promise<Delegate[]> {
...
-export function useDelegates(page: number, pageSize: number): DelegatesHookResponse {
+// TODO: Remove this eslint-disable when we have a proper filter implementation
+/* eslint-disable @typescript-eslint/no-unused-vars */
+export function useDelegates(
+  pagination: PaginationOption,
+  filters?: FilterOption<DelegateFilter>[],
+  sorting?: SortingOption<DelegateSorting>
+): DelegatesHookResponse {
+  const { page, pageSize } = pagination;
   const { data, error, mutate, isValidating } = useSWR<Delegate[]>(`delegates-${page}-${pageSize}`, () =>
```
rafinskipg, line 56: need to add the key on the useSWR , composed by the new filters, otherwise the data would not be correctly cached

apps/webapp/src/modules/delegates/hooks/useDelegatesFilters.tsx
```diff
@@ -0,0 +1,74 @@
...
+  };
+
+  useEffect(() => {
+    const newFilters: FilterOption<DelegateFilter>[] = [];
+
+    if (debouncedName) {
+      newFilters.push({
+        field: 'name',
+        value: debouncedName,
+        logicOperation: '$LIKE'
```
rafinskipg, line 38: is there any scenario where we are going to use anything except the $LIKE filter?

I don't think our UI will have things like filter by delegates with less than 50 MKR or things like that. 

Anyway, i think it should be supported on the useDelegates() since that is application agnostic, and anyone should be able to useDelegates() and filter by whatever. 

I have not thought about other ways of doing this system, so it seems good to me right now

rafinskipg (changes requested):

packages/hooks/src/delegates/useDelegates.ts
torhector2: Yes, I decided to not do any changes inside the hook until the subgraph was ready, but I can change that now.

packages/hooks/src/delegates/delegates.d.ts
```diff
@@ -46,3 +46,15 @@ export type DelegatesHookResponse = MakerReadHook & {
 export type DelegateHookResponse = MakerReadHook & {
   data?: Delegate;
 };
+
+export type DelegateFilter = {
+  address?: string;
```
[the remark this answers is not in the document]
torhector2: sounds good 👍. I'll change it.

apps/webapp/src/modules/delegates/hooks/useDelegatesFilters.tsx
```diff
@@ -0,0 +1,60 @@
...
+    setSorting({ field: sorting.field, order });
+  };
+
+  useEffect(() => {
+    const newFilters: FilterOption<DelegateFilter>[] = [];
+
+    if (debouncedText) {
+      newFilters.push(createFilter('id', debouncedText));
+    }
+
```
torhector2, line 35: These missing filters should be property names of the subgraph entity: Delegate

These will be the filters we pass down to the useDelegates hook.

apps/webapp/src/modules/delegates/types.ts
```diff
@@ -0,0 +1,18 @@
+// export type DelegateFilter = {
+//   text?: string;
+//   strategies?: string[];
+//   dvc?: string;
+// };
+
+export type DelegateFilter = {
```
torhector2, line 7: These missing filters bellow should be property names of the subgraph entity: Delegate

packages/hooks/src/delegates/useDelegates.ts
```diff
@@ -44,9 +55,16 @@ function fetchDelegates(page = 0, pageSize = 10): Promise<Delegate[]> {
...
 }
 
-export function useDelegates(page: number, pageSize: number): DelegatesHookResponse {
-  const { data, error, mutate, isValidating } = useSWR<Delegate[]>(`delegates-${page}-${pageSize}`, () =>
-    fetchDelegates(page, pageSize)
+// TODO: maybe we could get the graph types and force T to extend the Delegate entity from the subgraph
+// T extends DelegateEntity (or sth like that)
+export function useDelegates<T, U>(
+  pagination: PaginationOption,
+  filters?: FilterOption<T>[],
```
rafinskipg, line 62: as we talked , maybe instead of definning a full system capable of many things but inherently complex, we can just define the main characteristics delegates can be filtered by, and use that as a typed object that can be easily understood by developers/users

```
type DelegateFilters = {
  text?: string,
  delegators: FilterAmount,
  mkrDelegated: FilterAmount,
  createdAt: FilterAmount,
 }

{
  delegators: { 
     gte: 5,
   }, 
   mkrBalance: {
    lte: 1000000
    },
    text: 'Pepe'
    }  
```

rafinskipg (approved):

PR #76, opened 2023-03-16
add hooks to read and write vault contracts

b-pmcg:
### What does this PR do?
- set up contracts/abi boilerplate for vault-related contracts
- add hooks for reading vault-related params
- use math pkg to calculate values
- use some string formatting for human readable values

Screen Shot 2023-03-16 at 5 06 29 PM


packages/hooks/src/vaults/useVault.ts
```diff
@@ -0,0 +1,227 @@
+import { useContractRead, useContractReads } from 'wagmi';
+import { contracts, abis } from '@makerdao-dux/contracts';
+import {
```
rafinskipg, line 3: Maybe split this file into 3 different hooks files? 
useIlk
useUserVaults
useVault(id)

b-pmcg: yes, these hooks are just a work in progress, once I get everything working correctly I'll design the hooks better

packages/hooks/src/vaults/useVault.ts
```diff
@@ -0,0 +1,183 @@
...
+    select(data) {
+      const { ilks, ids, urns } = data;
+      const vaultsOwned = ids.map((id, i) => {
+        return { id, ilk: ilks[i], urn: urns[i] };
+      });
+      return vaultsOwned;
+    }
+  });
+
+  return { data, isLoading, error, mutate: refetch };
```
rafinskipg, line 76: The useSomething that read from contracts should export the Datasources definitions

apps/webapp/src/pages/Account.tsx
```diff
@@ -1,14 +1,118 @@
...
+    </Card>
+  );
+}
+
+type VaultProps = {
+  address: Address;
+};
+
+function VaultInfo({ address }: VaultProps): React.ReactElement {
+  const { data: vaults } = useGetCdps(address);
```
rafinskipg, line 88: should we call it useUserVaults ?

packages/utils/src/index.ts
```diff
@@ -1,3 +1,17 @@
 export { getEtherscanLink } from './getEtherscanLink';
 export { formatAddress } from './formatAddress';
 export { chainId } from './chainId';
+export {
+  annualStabilityFee,
```
rafinskipg, line 5: maybe we can export under a "math" namespace? or under a "vault/math" 

I guess all this math is only related to Vaults. Otherwise the package has lots of exported functions

packages/hooks/src/vaults/useVault.ts
```diff
@@ -0,0 +1,183 @@
...
+    args: [ilk],
+    scopeKey: `spot-ilks-${ilk}`
+  });
+
+  return { data, isLoading, error, mutate: refetch };
+}
+
+// TODO update with cdpmanager.ilks & urns hook so we don't have to pass in ilk/urn
+export function useVault(id: BigNumber, ilk: `0x${string}`, urn: Address): any {
+  const { data: vatIlkData } = useVatIlk(ilk);
```
rafinskipg, line 167: do you think the endUsers (devs) need to use the "useVatIlk" and other hooks defined here?

I'm thinking in terms of developer experience of the hooks package. The premise of making it easy to use. But i don't know about the vatILK maybe is data that is needed in other cases, so it makes sense to export as hook. 

We should add some documentation about these hooks and split them into different folders to keep them organized, if they are useful to be exported

torhector2 (commented):
LGTM so far

apps/webapp/src/pages/Account.tsx
```diff
@@ -1,14 +1,118 @@
...
+
+  return (
+    <>
+      <Heading size="sm">{numVaults || '0'} Vaults found for Proxy Address</Heading>
+      <Text>{address}</Text>
+      {vaults && Array.isArray(vaults) && (
+        <Flex sx={{ gap: 3 }}>
+          {vaults.map((vault, i) => {
+            return (
+              <Flex key={i}>
```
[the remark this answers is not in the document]
b-pmcg: yeah I agree. I removed it temporarily to make prototyping easier, but I will make sure to add the datasources when it's finished.

apps/webapp/src/pages/Account.tsx
b-pmcg: GetCdps is the actual name of the contract, I was trying to be consistent with naming the base hooks after the contract they call. But I think I'll wrap it in a more user friendly named hook like you suggest.

packages/hooks/src/vaults/useVault.ts
b-pmcg: Yeah, my goal is to have full coverage of the necessary contract functions as hooks as a base layer, and then more user friendly hooks which call multiple functions and parse the data to be better for devs. The base layer hooks probably do not need to be exported (eg `useVatIlk`) but the higher-level, user-friendly hooks like `useVault` will be exported.

Optionally we could export the base hooks in a different namespace to give full control to advanced devs, but generally I agree we shouldn't confuse people. And above all everything should be well documented, I'll make sure to add to the README before I'm finished.

rafinskipg: nice @b-pmcg  we used chat-gpt to generate the docs for the hooks, there is one page for each hook documented. 

Let me know if you need some info for that, otherwise we can do it later

packages/hooks/src/vaults/useVault.ts
```diff
@@ -0,0 +1,271 @@
...
+    ? `vault-info-${spot.toString()}-${rate.toString()}-${ink.toString()}-${art.toString()}-${par.toString()}-${mat.toString()}`
+    : null;
+
+  const {
+    data,
+    isValidating: isLoadingData,
+    error: errorData
+  } = useSWR(keyCalculateVaultInfo, () =>
+    allLoaded ? calculateVaultInfo({ spot, rate, art, ink, par, mat }) : undefined
+  );
```
b-pmcg, line 63: @rafinskipg idk if we need useSWR here since all the hooks can be cached by wagmi useContractRead. And we can pass the loading & error states straight from wagmi too, like how I did with useCollateralData below.

rafinskipg: yeah you are right!

packages/hooks/src/vaults/useCreateVault.ts
```diff
@@ -0,0 +1,57 @@
...
+  const acct = connectedAddress || ZERO_ADDRESS;
+  const dsProxy = useDsProxyData(acct);
+
+  // encode deposit function call for dsProxy to execute
+  const proxyActionsIface = new ethers.utils.Interface(contracts[chainId]?.PROXY_ACTIONS_DSR?.abi);
+  const encodedOpenVaultData = proxyActionsIface.encodeFunctionData('open', [ilk, acct]);
+
+  // prepare tx config
+  const { config: openVaultConfig, refetch } = usePrepareContractWrite({
+    address: dsProxy?.data?.dsProxyAddress,
```
b-pmcg, line 33: @rafinskipg 
address should be the PROXY_ACTIONS address (mainnet: `0x82ecD135Dce65Fbc6DbdD0e4237E0AF93FFD5038 `). 

Abi is the PROXY_ACTIONS abi, find it here: https://chainlog.makerdao.com/

functionName should be any of the ones on the contract, seen here: https://github.com/makerdao/dss-proxy-actions

args for each function are documented above too. I don't think you need to use the ethers `Interface` or encode the data

adamgoth: `PROXY_ACTIONS_DSR` is a contract specific to DSR functionality: https://etherscan.io/address/0x07ee93aeea0a36fff2a9b95dd22bd6049ee54f26#writeContract

For vaults, you want `PROXY_ACTIONS`: https://etherscan.io/address/0x82ecd135dce65fbc6dbdd0e4237e0af93ffd5038#writeContract

adamgoth: > args for each function are documented above too. For example with `open`, the final argument is `usr` which should be the users proxy address.
> 
> I don't think you need to use the ethers `Interface` or encode the data

I believe the DSR proxy actions require the function to be called by the proxy contract (rather than directly by the owner), but this may not be true for vault proxy actions. Taking a look now.

adamgoth: According to the technical docs, it should be used by the ds proxy and not by the user directly: https://docs.makerdao.com/smart-contract-modules/proxy-module/proxy-actions-detailed-documentation#3.-key-mechanisms-and-concepts

This part has a specific example of opening the vault with the proxy actions: https://docs.makerdao.com/smart-contract-modules/proxy-module/proxy-actions-detailed-documentation#proxy-action-usage-example-how-a-proxy-call-can-look-like

rafinskipg: Would this be enough? https://github.com/makerdao-dux/metadao-frontend/pull/84

adamgoth: > Would this be enough? #84

It looks right to me

apps/webapp/src/pages/Account.tsx
```diff
@@ -1,14 +1,175 @@
...
 import { ConnectButton } from '@rainbow-me/rainbowkit';
+import { useParams } from 'react-router-dom';
+import { utils } from 'ethers';
+import { formatAmount, formatPercent } from '@makerdao-dux/utils';
+
+const formatWad = (value?: bigint) => (value ? parseFloat(utils.formatUnits(value)) : 0);
+const formatRay = (value?: bigint) => (value ? parseFloat(utils.formatUnits(value, 27)) : 0);
+const formatRad = (value?: bigint) => (value ? parseFloat(utils.formatUnits(value, 45)) : 0);
+
+function VaultCard({ vault }: { vault: VaultIdentifiers }): React.ReactElement {
```
torhector2, line 38: I'd move VaultCard and other vaults components to separate files.

apps/webapp/src/pages/Vaults.tsx
```diff
@@ -0,0 +1,92 @@
...
+          <Heading size="sm">Vault #{vaultId}</Heading>
+          <Heading size="sm">{vaultData.collateralType}</Heading>
+        </Flex>
+      </CardHeader>
+      <CardBody>
+        <DataSourceInfo external={false} sources={dataSources}>
+          <Stack>
+            <Flex sx={{ justifyContent: 'space-between' }}>
+              <Text>Delayed Price</Text>
+              <Text>{formatAmount.format(formatWad(vaultData?.delayedPrice))}</Text>
```
torhector2, line 28: What about create util functions for these formats?
instead of
`formatAmount.format(formatWad(vaultData?.delayedPrice))`


`formatAmount(vaultData?.delayedPrice, 'wad')`

where the old formatAmount is renamed to AmountNumberFormat and we create a formatAmount function

```
function formatAmount(amount: number, unit:'wad' | 'ray' | 'rad') {
  // pick the right method based on unit
  return amountNumberFormat.format(formatWad(vaultData?.delayedPrice))
}
```

or we can have 3 functions:
```
formatWadAmount(amount)
formatRadAmount(amount)
formatRayAmount(amount)
```

apps/webapp/src/pages/Vaults.tsx
```diff
@@ -0,0 +1,92 @@
...
+          </Stack>
+        </DataSourceInfo>
+      </CardBody>
+    </Card>
+  ) : (
+    <>No vault found with ID {vaultId}</>
+  );
+}
+
+function UserVaultsContainer({ address }: { address: Address }): React.ReactElement {
```
torhector2, line 47: same, I'd move these private components into separate files

packages/hooks/src/vaults/useCollateralTypes.tsx
```diff
@@ -0,0 +1,31 @@
+import { MakerReadHook } from '../hooks';
+import { SupportedCollateralTypes } from './vaults.constants';
+import { CollateralType } from './vault';
+
+export function useCollateralTypes(): MakerReadHook & {
```
torhector2, line 5: is a bit weird to type the response like this here when we're just returning a const array and the rest of the props are not being used.

torhector2 (commented):
Great work :D just some minor comments.

packages/hooks/src/vaults/useVault.ts
```diff
@@ -286,26 +299,27 @@ function useSpotPar(): UseSpotParHookResponse {
...
+            onChain: true,
+            href: getEtherscanLink(chainId, contractAddress, 'address'),
+            trustLevel: TRUST_LEVELS[0]
+          }
+        ]
+      : []
   };
 }
 
 type JugIlkData = ReadContractResult<typeof mcdJugABI, 'ilks'>;
```
rafinskipg, line 315: i think we can get rid of this

apps/webapp/src/pages/Vaults.tsx
b-pmcg: good idea, implemented.

apps/webapp/src/pages/Vaults.tsx
b-pmcg: Normally I'd agree, but this page isn't really the final design, and since we're halting work on vaults let's just wait until we restart that work to nail down the component design.

packages/hooks/src/vaults/useCollateralTypes.tsx
rafinskipg: is a mock for now, until we get the real collateral types

packages/utils/src/formatUnits.ts
```diff
@@ -0,0 +1,9 @@
+import { formatUnits, parseUnits } from 'ethers';
+
+export const formatWad = (value?: bigint) => (value ? parseFloat(formatUnits(value)) : 0);
+export const formatRay = (value?: bigint) => (value ? parseFloat(formatUnits(value, 27)) : 0);
+export const formatRad = (value?: bigint) => (value ? parseFloat(formatUnits(value, 45)) : 0);
+
```
rafinskipg, line 6: maybe we should add some inline comments with links to this: https://github.com/makerdao/dss/blob/master/DEVELOPING.md#units

packages/utils/src/formatValue.ts
```diff
@@ -0,0 +1,28 @@
...
+  maximumFractionDigits: 2
+});
+
+export function formatAmount(amount: bigint, unit: 'wad' | 'ray' | 'rad' = 'wad') {
+  const amountToFormat =
+    unit === 'ray' ? formatRay(amount) : unit === 'rad' ? formatRad(amount) : formatWad(amount);
+  return formatUnitAmount.format(amountToFormat);
+}
+
+export function formatPercent(amount: bigint, unit: 'wad' | 'ray' | 'rad' = 'wad') {
```
rafinskipg, line 21: maybe we could add some unit tests for the format functions

rafinskipg (approved):
LGTM, 2 comments regarding documenting a bit the format functions and adding unit tests

torhector2 (approved):
LGTM 💯

packages/utils/src/formatValue.ts
b-pmcg: good call, especially since these will be used beyond just the vault hooks code

PR #79, opened 2023-03-24
Add collateral types

rafinskipg:
Refactor the exports of the hooks to organize

Updated the concept of "collateral types" to be able to render the page with the allowed vault types


packages/hooks/src/vaults/useCollateralTypes.tsx
```diff
@@ -0,0 +1,34 @@
...
+export function useCollateralTypes(): MakerReadHook & {
+  data?: CollateralType[];
+} {
+  return {
+    data: [
+      {
+        id: SupportedCollateralTypes.ETH_A,
+        name: 'ETH-A',
+        symbol: 'ETH',
+        minCollateralizationRatio: 0.5
```
b-pmcg, line 14: I don't think we should hardcode `minCollateralizationRatio` because it is a risk param that comes from the chain and updated by governance

b-pmcg (approved):

PR #80, opened 2023-03-27
Update delegates query subgraph

torhector2:
Updates useDelegates and the search functionality to use the new queries and filters


apps/webapp/src/modules/search/hooks/useSearchGovernanceData.tsx
```diff
@@ -5,11 +5,14 @@ import { searchGovernanceData, SearchResult } from '../utils/searchGovernanceDat
 
 export function useSearchGovernanceData(query: string) {
   const [results, setResults] = useState<SearchResult[] | undefined>();
-  const { data: delegates, isLoading: isLoadingDelegates } = useDelegates({ page: 0, pageSize: 999 });
+  const debouncedQuery = useDebounceQuery(query, query === '' ? 0 : 500);
+  const { data: delegates, isLoading: isLoadingDelegates } = useDelegates(
+    { page: 0, pageSize: 999 },
```
rafinskipg, line 10: i think now we can remove the pagesize, because we can search using the graph search queries

rafinskipg (approved):

PR #82, opened 2023-03-27
Update search subgraph

torhector2:
- Updates search function with subgraph query
- Update voting committees fetcher function with subgraph query
- Add strategies fetcher function using subgraph

NOTE: Shall we move `useSearchGovernanceData` and fetcher to the **hooks** package and leave only `useGlobalSearch` hook in the **webapp** package?


apps/webapp/src/modules/search/hooks/useSearchGovernanceData.tsx
```diff
@@ -1,41 +1,134 @@
...
+  }, [isLoading]);
 
-      if (debouncedQuery === '') {
-        setResults(undefined);
-      } else {
-        searchFn();
+  return { results: data, isLoading: displayLoading };
+}
+
+// TODO: ALL this may go to hooks package
```
rafinskipg, line 39: agree, might go to the hooks package

apps/webapp/src/modules/search/hooks/useSearchGovernanceData.tsx
```diff
@@ -1,41 +1,134 @@
...
+
+  return request<SearchDataSourceResponse>(URL_SUBGRAPH, query).then(data => {
+    const delegates = data.delegates.map(mapDelegates);
+    const dvcs = data.votingCommittees.map(mapVotingCommittees);
+    const strategies = data.votingStrategies.map(mapVotingStrategies);
+    return [...delegates, ...strategies, ...dvcs];
+  });
+}
+
+function mapDelegates(delegate: Delegate) {
```
rafinskipg, line 100: if this goes to the hook package, the mapping should probably remain in the app

packages/hooks/src/delegateVotingCommittees/useDelegateVotingCommittees.ts
```diff
@@ -1,16 +1,59 @@
...
+          description
+          id
+          name
+        }
+      }
+    }
+  `;
+
+  return request(URL_SUBGRAPH, query).then(data => {
+    return data.votingCommittees.map((dvc: DelegateVotingCommittee) => {
```
rafinskipg, line 53: unnecesary map

packages/hooks/src/delegateVotingCommittees/useDelegateVotingStrategies.ts
```diff
@@ -0,0 +1,51 @@
...
+        description
+        id
+        name
+      }
+    }
+  `;
+
+  return request(URL_SUBGRAPH, query).then(data => {
+    return data.votingStrategies.map((strategy: Strategy) => {
+      return {
```
rafinskipg, line 46: same

rafinskipg (approved):

packages/hooks/src/searchGovernanceData/useSearchGovernanceData.tsx
```diff
@@ -0,0 +1,61 @@
+import request, { gql } from 'graphql-request';
+import useSWR from 'swr';
+import { URL_SUBGRAPH } from '../constants';
+import { GovernanceData } from './governanceData';
+
+export function useSearchGovernanceData(query: string) {
```
rafinskipg, line 6: you need to define the type

MakerReadHook & { data : Whatever }

rafinskipg (changes requested):

rafinskipg (approved):

PR #84, opened 2023-03-28
Add PROXY_ACTIONS

rafinskipg:
Updates the create vault hook to use PROXY_ACTIONS instead of PROXY_ACTIONS_DSR


torhector2 (approved):
LGTM

PR #85, opened 2023-03-28
Wagmi cli

b-pmcg:
### What does this PR do?
- uses wagmi cli to generate abis and address config helpers
- replaces instances of using contracts constant with the generated ones


apps/webapp/package.json
```diff
@@ -52,7 +52,7 @@
   "peerDependencies": {
     "ethers": "^5.7.2",
     "react": "^18.2.0",
-    "wagmi": "^0.10.11"
+    "wagmi": "^0.12.7"
```
rafinskipg, line 55: this has been updated in another PR, the one from hector, we updated the wagmi deps

packages/hooks/src/dsProxy/useDsProxyBuild.ts
```diff
@@ -1,15 +1,17 @@
...
 
 export function useDsProxyBuild({ onSuccess, onError }: MakerWriteHookParams): MakerWriteHook {
   const chainId = useChainId();
 
+  const { abi, address } = proxyRegistryConfig;
+
   const { config, refetch } = usePrepareContractWrite({
-    address: contracts[chainId]?.PROXY_REGISTRY?.address,
-    abi: contracts[chainId]?.PROXY_REGISTRY?.abi,
+    address: address[chainId as keyof typeof address],
```
rafinskipg, line 13: we need the types to be exported correctly without having to do keyof typeof,

packages/hooks/src/dsProxy/useDsProxyData.ts
```diff
@@ -14,24 +14,24 @@ export type DsProxyHookResponse = MakerReadHook & {
 };
 
 export function useDsProxyData(owner?: `0x${string}`): DsProxyHookResponse {
-  const chainId = useChainId();
+  const chainId = useChainId() as keyof typeof contractAddress;
   const { address: connectedAddress } = useAccount();
 
-  const contractAddress = contracts[chainId]?.PROXY_REGISTRY?.address;
+  const { address: contractAddress, abi } = proxyRegistryConfig;
```
rafinskipg, line 20: can we have the contracts grouped by chain like before?

packages/hooks/src/dsProxy/useDsProxyData.ts
```diff
@@ -14,24 +14,24 @@ export type DsProxyHookResponse = MakerReadHook & {
 };
 
 export function useDsProxyData(owner?: `0x${string}`): DsProxyHookResponse {
-  const chainId = useChainId();
+  const chainId = useChainId() as keyof typeof contractAddress;
```
rafinskipg, line 17: also, we shouldn't need to do this

rafinskipg (changes requested):
We need to find a better way to interface with the contracts so all the types are exported correctly. 

- Group by chain (idk if that is already the case)
- Remove all the old contracts and references

packages/hooks/src/dsProxy/useDsProxyData.ts
b-pmcg: yeah it sucks :(

packages/hooks/src/dsProxy/useDsProxyBuild.ts
b-pmcg: This was because the keys of the address object are readonly and we could not use `chainId`, a number to key off them. I added `parseConfigForChain` which parses the address correctly without using `keyof typeof`.

packages/hooks/src/dsProxy/useDsProxyData.ts
b-pmcg: We can now get the contract by chain ID by using either `getAddressByChainId(addressGroup, chainId)` or `parseConfigForChain(contractConfig, chainId)` depending on the need. The generated file groups chainId's within the address object rather than the other way around.

packages/contracts/src/index.ts
```diff
@@ -1,47 +1,31 @@
...
-  WSTETH = 'WSTETH',
-  OP = 'OP'
+  PIP_RETH = 'PIP_RETH'
 }
 
-export type Contract = {
-  address: `0x${string}`;
-  abi: any;
+export const getAddressByChainId = (addressGroup: Record<number, string>, chainId: number) => {
+  if (Object.hasOwn(addressGroup, chainId)) {
```
rafinskipg, line 20: I don't think we need any of these helpers, if addressGroup is already an object with chainId inside we can access directly

packages/hooks/src/dsProxy/useDsProxyData.ts
```diff
@@ -17,9 +17,11 @@ export function useDsProxyData(owner?: `0x${string}`): DsProxyHookResponse {
   const chainId = useChainId();
   const { address: connectedAddress } = useAccount();
 
-  const contractAddress = contracts[chainId]?.PROXY_REGISTRY?.address;
+  const { address: contractAddresses, abi } = proxyRegistryConfig;
 
-  const acct = owner || connectedAddress;
+  const contractAddress = getAddressByChainId(contractAddresses, chainId);
```
rafinskipg, line 22: so we could just do `contractAddresses[chainId]` ? I thinkt hat would be more neat

packages/hooks/src/dsr/useChi.ts
```diff
@@ -1,19 +1,19 @@
...
 import { useChainId } from '../network/useChainId';
 import { getEtherscanLink } from '@makerdao-dux/utils';
+import { parseConfigForChain, mcdPotConfig } from '@makerdao-dux/contracts';
 
 type UseChiHookResponse = MakerReadHook & { data?: BigNumber };
 export function useChi(): UseChiHookResponse {
   const chainId = useChainId();
 
-  const contractAddress = contracts[chainId]?.MCD_POT?.address;
+  const { address: contractAddress, abi } = parseConfigForChain(mcdPotConfig, chainId);
```
rafinskipg, line 12: same `mcdPotConfig.address[chainId]` would be enough right? no need to import another function to access a contract information. 

I think the contracts package should export things easily and ready to use.

rafinskipg (changes requested):

packages/contracts/src/index.ts
torhector2: Although I agree with what Rafa says in terms of usability, I can see it doesn't compile due to the readonly-ness of the generated types. We can try to find a different approach to export the generated types or use the helpers.

packages/hooks/src/dsr/useChi.ts
torhector2: In order to work you need to access like this :S

`mcdPotConfig.address[chainId as 1 | 5];`

adamgoth:
> When I connect my wallet on the DSR page, I get an error that crashes the page. The error message seems be "execution reverted" which isn't super helpful. I'll try to do a little more debugging locally.

It only appears to happen on goerli and not mainnet

packages/hooks/src/dsProxy/useDsProxyBuild.ts
```diff
@@ -1,17 +1,10 @@
...
 
 export function useDsProxyBuild({ onSuccess, onError }: MakerWriteHookParams): MakerWriteHook {
-  const chainId = useChainId();
-
-  const { abi, address } = parseConfigForChain(proxyRegistryConfig, chainId);
-
-  const { config, refetch } = usePrepareContractWrite({
-    address,
-    abi,
+  const { config, refetch } = usePrepareProxyRegistryWrite({
```
rafinskipg, line 7: love this

packages/hooks/src/dsProxy/useDsProxyData.ts
```diff
@@ -17,24 +17,16 @@ export function useDsProxyData(owner?: `0x${string}`): DsProxyHookResponse {
...
     error,
     refetch
-  } = useContractRead({
-    address: contractAddress,
-    abi,
-    functionName: 'proxies',
-    args: [acct],
-    enabled: acct !== ZERO_ADDRESS
-  });
+  } = useProxyRegistryRead({ functionName: 'proxies', args: [acct], enabled: acct !== ZERO_ADDRESS });
```
rafinskipg, line 29: nice

packages/hooks/src/dsr/useChi.ts
```diff
@@ -1,21 +1,16 @@
...
 import { getEtherscanLink } from '@makerdao-dux/utils';
-import { parseConfigForChain, mcdPotConfig } from '@makerdao-dux/contracts';
+import { useMcdPotRead, useMcdPot } from '@makerdao-dux/contracts';
 
 type UseChiHookResponse = MakerReadHook & { data?: BigNumber };
 export function useChi(): UseChiHookResponse {
   const chainId = useChainId();
 
-  const { address: contractAddress, abi } = parseConfigForChain(mcdPotConfig, chainId);
+  const { address: contractAddress } = useMcdPot();
```
rafinskipg, line 11: so clean

packages/hooks/src/dsr/usePie.ts
```diff
@@ -10,13 +9,8 @@ type UsePieHookResponse = MakerReadHook & { data?: BigNumber };
...
 
-  const { address: contractAddress, abi } = parseConfigForChain(mcdPotConfig, chainId);
-
-  const { data, isLoading, error, refetch } = useContractRead({
-    address: contractAddress,
-    abi,
-    functionName: 'Pie'
-  });
+  const { address: contractAddress } = useMcdPot();
+  const { data, isLoading, error, refetch } = useMcdPotRead({ functionName: 'Pie' });
```
rafinskipg, line 13: nice

packages/hooks/src/network/useChainId.ts
```diff
@@ -1,10 +1,14 @@
...
-export function useChainId(): number {
+export function useChainId() {
   const network = useNetwork();
 
   const chainId = network.chain?.id || 1;
 
-  return chainId;
+  const ret = { chainId };
+
+  Object.freeze(ret);
```
rafinskipg, line 11: i don't understand why we need this?

rafinskipg (approved):
This looks neat 

I pretty much approve this approach

PR #86, opened 2023-03-29
Randomize delegates on the frontend

torhector2:
Randomize the delegates list on the frontend by randomly selecting a query sort of the following fields:

`'totalDelegated', 'delegators', 'metadata.name',  'ownerAddress', 'txnHash', 'blockNumber'`
and combining them with a random sort order `'asc', 'desc'`

The randomization occurs only when the page loads, the user selects "Default" or the user press the Reset button.

image


apps/webapp/src/modules/delegates/hooks/useDelegatesFilters.tsx
```diff
@@ -10,19 +22,22 @@ export const useDelegatesFilters = () => {
   const debouncedText = useDebounceQuery(text, 300);
   const debouncedDvc = useDebounceQuery(dvc, 300);
 
+  const randomizeSorting = () => {
+    const field = getRandomItem(delegateSortFields) as DelegateSort['field'];
+    const order = getRandomItem(orderDirection) as DelegateSort['order'];
+
+    setSortingValue(field, order);
+  };
```
adamgoth, line 30: Seems to work well, but I wonder if it'll appear to be "random" when the "random sort" ends up as total delegated desc, or alphabetical by name, or by blockNumber (creation date) etc.

Is there a reason we can't just randomly shuffle the array with some algorithm based on Math.random each time instead?

torhector2: Right, the problem was that by doing that, we'd end up losing the subgraph pagination. Maybe we can try to shuffle the content of each page instead, but we should be careful as each re-render may reshuffle the previous pages.

adamgoth: Ah right, makes sense. In that case, I think your solution serves the purpose well 🤝

adamgoth (approved):

apps/webapp/src/modules/delegates/hooks/useDelegatesFilters.tsx
rafinskipg: i agree with reshufling the content of the page could be the right solution to implement partial randomnes, the order could be kept by a wrapper hook that keeps a map (key, value) of the order of each page, 

BUT! I also think that this is not really important right now,

rafinskipg (approved):

PR #87, opened 2023-03-30
Randomize delegates list pages for default sorting

torhector2:
Randomize the delegates pages when the sorting option is "Default".


adamgoth (approved):
Seems to work well for me 💯

rafinskipg (approved):

PR #88, opened 2023-03-30
ChainId

rafinskipg:
Remove the helper function & handle invalid chains


packages/hooks/src/dsProxy/useDsProxyBuild.ts
```diff
@@ -1,17 +1,22 @@
...
 import { onSettledWrapper } from '../helpers/wagmiUseContractWriteParameters';
+import { INVALID_CHAIN_WRITE_HOOK_RESPONSE } from '../constants';
 
 export function useDsProxyBuild({ onSuccess, onError }: MakerWriteHookParams): MakerWriteHook {
   const chainId = useChainId();
 
-  const { abi, address } = parseConfigForChain(proxyRegistryConfig, chainId);
+  if (chainId !== mainnet.id && chainId !== goerli.id) {
+    return INVALID_CHAIN_WRITE_HOOK_RESPONSE;
+  }
```
adamgoth, line 13: I think this should be a helper function so that it only needs to be updated in one place if we add new chains

adamgoth (changes requested):
Looks like the build isn't working. Are you still exporting this somewhere?

Screenshot 2023-03-30 at 3 50 42 PM

packages/hooks/src/dsProxy/useDsProxyBuild.ts
rafinskipg: the problem is the typing, try downloading the branch and moving the function to a helper

rafinskipg:
> Looks like the build isn't working. Are you still exporting this somewhere?
> 
> Screenshot 2023-03-30 at 3 50 42 PM

fixed

packages/hooks/src/dsProxy/useDsProxyBuild.ts
torhector2: Yes this one is a tricky one, not sure if it'd work but we could try something like type predicates for a helper. I didn't really think it through but I may worth give it a try.

Specially if this is the only check `(chainId !== mainnet.id && chainId !== goerli.id)` as it's more predictable. I though we would have other chainId combinations with Opt and Arb.

https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates

torhector2 (commented):
I like the approach, I'd give it a try to the type predicates I mentioned above.

b-pmcg:
Personally I don’t think this fix should go into the wagmi-cli PR because I think it addresses a separate problem, which is how to handle unsupported chains. I don’t think the hook should have the logic of determining whether a chain is supported or not, it should just do the job of calling the contract.

The issue with unsupported chains already existed in the main branch, with the old `contracts` object, if you use an unsupported ID, it would silently return an undefined address & abi.

The reason I added the helper functions in the other PR is to solve the specific problem with the generated address keys being readonly, and the `chainId` returned by `useChainId` (via wagmi’s `useNetwork()` hook) is not able to key off the generated address object.

With this solution we can never use any other network than goerli or mainnet, try adding `&& chainId !== arbitrum.id` to the invalid hook conditional and you’ll see the error `Property '42161' does not exist on type...` because we would need to add arbitrum addresses for every contract to make the compiler happy.

PR #89, opened 2023-04-03
Basic layout for detail pages

torhector2:
Creates a basic layout for detail pages.

Desktop
image

Mobile
image


rafinskipg:
@torhector2 mhh i don't know if we need a generic layout for detail pages, is this repeated through the app? 

otherwise i think just define the styles on the delegate page

torhector2:
@rafinskipg there was this ticket where it was suggested that
https://app.shortcut.com/dux-makerdao/story/1991/create-a-layout-component-for-the-different-profile-pages

apps/webapp/src/modules/layout/components/detail-layout/LeftSection.tsx
```diff
@@ -0,0 +1,16 @@
+import { Box, SystemStyleObject } from '@chakra-ui/react';
+
+interface LeftSectionProps {
+  children: React.ReactElement;
+  sx?: SystemStyleObject;
+}
+
+export function LeftSection({ children, sx }: LeftSectionProps): React.ReactElement {
+  return (
```
rafinskipg, line 9: i don't think we need to create this component, is just a div, better just leave it on the detaillayout

apps/webapp/src/modules/layout/components/detail-layout/MainSection.tsx
```diff
@@ -0,0 +1,16 @@
+import { Box, SystemStyleObject } from '@chakra-ui/react';
+
+interface MainSectionProps {
+  children: React.ReactElement;
+  sx?: SystemStyleObject;
+}
+
+export function MainSection({ children, sx }: MainSectionProps): React.ReactElement {
+  return (
+    <Box as="main" overflowY="auto" sx={sx}>
```
rafinskipg, line 10: i don't think we need to create this component, is just a div, better just leave it on the detaillayout

rafinskipg (changes requested):

apps/webapp/src/modules/layout/components/detail-layout/MainSection.tsx
torhector2: @rafinskipg made those changes

rafinskipg (approved):

PR #90, opened 2023-04-03
Remove extra hooks from useVault

rafinskipg:
Since now all the contract calls are already exported, do we need this?

Also i'm not sure about this PR. 

if we remove all this code, we almost should do the same for the DSR and all the usePie useSmallPie hooks, 

maybe it makes sense to delete it all, idk ,


b-pmcg:
yeah, I think this is a really good idea. I updated the Dsr related hooks to use the generated ones

PR #93, opened 2023-04-04
Hide delegation elements and add farming page

torhector2:
- create a new farming page (can be empty)
- repurpose landing page
- repurpose search bar: returns empty results until we decide what to do with it
- hide delegates
- hide CVCs
- hide delegate detail page

image
image
image
image


apps/webapp/src/modules/intro/Landing.tsx
```diff
@@ -1,39 +1,29 @@
-import { Flex, Box, Image, Heading, Text, Button, Link } from '@chakra-ui/react';
-import { ArrowForwardIcon } from '@chakra-ui/icons';
-import { Link as ReactRouterLink } from 'react-router-dom';
-import { IntroStep } from './LandingIntro';
+import { Flex, Box, Image, Heading, Text } from '@chakra-ui/react';
 
-export function Landing({ setStep }: { setStep: (step: IntroStep) => void }): React.ReactElement {
```
adamgoth, line 6: Hmm, I wonder if we should create a module called `egf` and move some of this code there to preserve it. Feels bad to just delete it. I guess we could always recover it from git though.

adamgoth (approved):

apps/webapp/src/modules/intro/Landing.tsx
torhector2: true, good point. Will do.

rafinskipg (approved):

PR #94, opened 2023-04-04
transactions handler

rafinskipg:
Update: PR adds rainbow wallets recent tx history support, and also fixes rendering issues when no wallet is connected

---

Adds new transactions provider that keeps track of all the webapp transactions and allows to access them as a list.

This is useful to create a list of "latest" transactions, in the future we can even persist this list on a localstorage for the users to recover their latest txs.

Change the hooks write parameters and return types to include a "onStart" that is called when the hash is received, and also to return the hash on the hook call.

image


apps/webapp/src/modules/transactions/components/TransactionsIndicator.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { useState } from 'react';
+import { TransactionsList } from './TransactionsList';
+
+export function TransactionsIndicator() {
+  const { transactions } = useTransactions();
+  const [isOpen, setIsOpen] = useState(false);
+
+  return (
+    <Box m={3}>
+      <div
```
adamgoth, line 21: `div` can be `Button` for better accessibility

packages/hooks/src/dsr/useDsrWithdraw.ts
```diff
@@ -43,7 +45,12 @@ export function useDsrWithdraw({
     args: [proxyActionsDsrAddress, encodedJoinData],
     enabled: !!dsProxy?.data?.dsProxyAddress
   });
-  const { writeAsync, isLoading, error, data } = useContractWrite(withdrawConfig);
+  const { writeAsync, isLoading, error, data } = useContractWrite({
```
adamgoth, line 48: Do you think we should create a custom hook that automatically adds `onStart` to be called on `useContractWrite`'s `onSuccess` callback?

My concern is if someone writes a new hook, they would have to know to add `onStart` to `onSuccess`. If we write a custom hook, we can require it.

apps/webapp/src/modules/transactions/components/TransactionsList.tsx
```diff
@@ -0,0 +1,35 @@
...
+            <Flex sx={{ justifyContent: 'space-between' }}>
+              <Text variant="secondary">Tx Hash: {formatAddress(transaction.hash)}</Text>
+
+              <Text variant="secondary">
+                Time: {formatDistance(new Date(transaction.date), new Date())} ago
+              </Text>
+            </Flex>
+            <Flex sx={{ justifyContent: 'space-between' }}>
+              <Text variant="secondary">Status: {transaction.status}</Text>
+              <Link href={getEtherscanLink(transaction.chainId, transaction.hash, 'tx')}>
```
adamgoth, line 26: Should add `isExternal` prop so it opens in a new tab

adamgoth (commented):
This will be nice to have. I was trying it out on goerli and it looks like it still shows pending as the status when txs fail.

Screenshot 2023-04-04 at 10 35 09 AM

packages/hooks/src/dsr/useDsrWithdraw.ts
adamgoth: What I mean is just wrapping the `useContractWrite` hook to basically calls it from that custom hook instead and exposes the same functions, but automatically adds `onStart` to their `onSuccess` callback before returning everything.

so something like `useContractWriteWrapper(config, onStart)`:

```
function useContractWriteWrapper(config, onStart, <...anything else we want to be required for each write hook>) {
  const { writeAsync, error, isLoading, data } = useContractWrite({
    ...config,
    onSuccess: (d: SendTransactionResult) => {
      onStart(d.hash);
    }
  });
  
  return {
    writeAsync,
    error,
    isLoading,
    data
  }

}

```

I'm not sure, maybe it's not a good idea. I'm just concerned that `onStart` has to be added manually each time to `onSuccess` someone creates a new hook if we want it to be tracked in the history. Would be better DX if the dev doesn't have to remember that.

rafinskipg: if the dev only uses our hooks it will have to add the onStart, or just not track any tx. 

if we want the devs to use any wagmi hooks, we should probably rename our hooks to not use an onstart, and have an onsucess and onSettled, and treat the onSuccess just a hash obtainer.  (so it matches the same or similar interface)

I'm not really in favour of the usecontractwritewrapper, seems not really necesary. 

Your concern is for the dev that has to write hooks in the hooks package or for the dev that has to use the hooks? 

or for the combination of being able to use any other hook?

adamgoth: I think you're right. The end user will still have to add `onStart` so it's fine. As the hooks lib authors, we'll just have to remember to add `onStart` to `onSuccess` callback anytime we use `useContractWrite` so that it gets tracked in history.

apps/webapp/src/modules/transactions/components/TransactionsList.tsx
```diff
@@ -0,0 +1,35 @@
...
+  return (
+    <Box>
+      {transactions.map(transaction => {
+        return (
+          <Box key={transaction.hash} sx={{ borderTop: '1px solid', paddingTop: 3, paddingBottom: 3 }}>
+            <Text variant="primary" sx={{ fontWeight: 'bold' }}>
+              {transaction.message}
+            </Text>
+            <Flex sx={{ justifyContent: 'space-between' }}>
+              <Text variant="secondary">Tx Hash: {formatAddress(transaction.hash)}</Text>
```
adamgoth, line 18: Tiago may want to offer some input here, but I think we could make the hash a link and add the external link icon and then not have to have text that says "View on Etherscan" on every history item

torhector2 (approved):
Very nice implementation 💯

adamgoth:
> This could also be worth a look: https://www.rainbowkit.com/docs/recent-transactions

Example of it:
Screenshot 2023-04-04 at 1 31 33 PM

PR #101, opened 2023-04-06
Up pnpm

torhector2:
- Force a version of pnpm >= 7.31
- Update lock file with dedup deps

This is what a user running pnpm install with a wrong version would see. In this example I force the version to be exactly 7.31 but I have v8.
image


adamgoth (commented):
💯 💯 nice addition

PR #104, opened 2023-04-12
Adding staking rewards contract and hooks

rafinskipg:
Added the mock contracts deployed at : 

From the test contracts: https://github.com/makerdao-dux/duxcoin-rewards

We have 
- a hook to get available farms (token / rewards)
- a hook to stake
- a hook to withdraw
- hooks for fetching balance and rewards from staking contracts

UI:
- A list of farms , using the useAvailableFarms hook
- A farmItem that calls the different approve/balance  /stake / withdraw/rewards functions
image


packages/hooks/src/staking/useStakeInTokenRewards.tsx
```diff
@@ -0,0 +1,62 @@
...
+  // const encodedOpenVaultData = proxyActionsContract.Interface.encodeFunctionData('stake', [
+  //   stakingRewardsAdress,
+  //   ilk,
+  //   dsProxyAddress
+  // ]);
+
+  // prepare tx config
+  const { config, refetch } = usePrepareStakingRewardsStake({
+    args: [amount],
+    enabled: !address || address === ZERO_ADDRESS || amount.isZero(),
```
torhector2, line 38: shouldn't this be the opposite? enabled when this condition is false 🤔

packages/hooks/src/staking/useWithdrawFromStakingRewards.tsx
```diff
@@ -0,0 +1,66 @@
...
+  // const encodedOpenVaultData = proxyActionsContract.Interface.encodeFunctionData('stake', [
+  //   stakingRewardsAdress,
+  //   ilk,
+  //   dsProxyAddress
+  // ]);
+
+  // prepare tx config
+  const { config, refetch } = usePrepareStakingRewardsWithdraw({
+    args: [amount],
+    enabled: !address || address === ZERO_ADDRESS || amount.isZero(),
```
torhector2, line 42: same here

packages/hooks/src/staking/useStakeInTokenRewards.tsx
b-pmcg: yep, that's right

packages/hooks/src/staking/useStakeInStakingRewards.tsx
```diff
@@ -1,45 +1,49 @@
...
-  //   ilk,
-  //   dsProxyAddress
-  // ]);
 
-  // prepare tx config
-  const { config, refetch } = usePrepareStakingRewardsStake({
+  const { config, refetch } = usePrepareContractWrite({
     args: [amount],
+    address: contractAddress,
+    abi: stakingRewardsDaiABI, // we should be able to use any staking rewards contract abi here
```
b-pmcg, line 32: yeah, as far as I know they will all use the same stakingRewards contract, so we can probably get away with that.

packages/hooks/src/staking/useStakeInStakingRewards.tsx
```diff
@@ -1,45 +1,49 @@
...
 import { BigNumber } from 'ethers';
+import { SendTransactionResult } from '@wagmi/core';
 
 type UseStakeInStakingRewardsResponse = MakerWriteHook & {
   execute: () => Promise<string | undefined>;
 };
 
+// Allows user to stake in a staking rewards contract
+// We need to provide the contract address of the staking rewards contract since there are many of them
 export function useStakeInStakingRewards({
```
b-pmcg, line 15: I'm not sure about this hook name, what about `useStakingRewardsStake`? Then we can have `useStakingRewardsClaim` & withdraw, etc.

rafinskipg: makes sense, i'll change it

apps/webapp/src/modules/tokens/types/token.d.ts
```diff
@@ -1,5 +1,5 @@
 export type Token = {
-  address?: string;
+  address?: `0x${string}`;
```
b-pmcg, line 2: I know we haven't really been consistent with this throughout the app, but I'm thinking we should use the `Address` type specifically for addresses, and `0x${string}` for any other hex string. Probably doesn't matter much, but it might help readability.

b-pmcg (approved):
lets go space farmer

PR #105, opened 2023-04-14
fix which modals to show, fix the user dsr balance calcs

b-pmcg:
### What does this PR do?
- fixes the conditionals to show modals on the DSR page
- adds new math functions to utils lib to calculate dsr data
- adds `exitAll` function to dsr withdraw to guarantee full withdrawl


apps/webapp/src/modules/dsr/components/DsrWithdrawModal.tsx
```diff
@@ -52,7 +53,7 @@ export function DsrWithdrawModal() {
         <Flex sx={{ flexDirection: 'column', maxWidth: 250 }}>
           <FormLabel>Withdraw Amount</FormLabel>
           <TokenInput
-            balance={dsrData?.userDsrBalance || BigNumber.from(0)}
+            balance={parseUnits(formatRay(dsrData?.userDsrBalance).toString()) || BigNumber.from(0)}
```
rafinskipg, line 56: why do we need to do this? what are the decimal places of userDsrBalance? is not 18?

rafinskipg (changes requested):
I think we should figure out how do we return numbers from the hooks package in the same way accross all the responses, so we avoid having a mix of bigints and BigNumbers

b-pmcg:
> I think we should figure out how do we return numbers from the hooks package in the same way accross all the responses, so we avoid having a mix of bigints and BigNumbers

@rafinskipg yeah good point, I think we should migrate away from ethers v5 since wagmi will soon be using bigints. I'll bring this up in the discord so all the devs can discuss.

apps/webapp/src/modules/dsr/components/DsrWithdrawModal.tsx
b-pmcg: yeah the dsrBalance is stored as a `ray` with 27 decimal places, so we have to convert it. When we are eventually able to move away from using bignumbers we won't have to do this conversion.

apps/webapp/src/modules/dsr/components/DsrDepositModal.tsx
```diff
@@ -18,30 +18,29 @@ import {
...
   const [depositAmount, setDepositAmount] = useState(BigNumber.from(0));
 
   // Allowance
   const [hasAllowance, setHasAllowance] = useState<boolean>(false);
 
   useEffect(() => {
-    if (allowance && depositAmount.gt(0)) {
-      setHasAllowance(BigNumber.from(allowance).gte(depositAmount));
+    if (allowance && allowance.gte(depositAmount)) {
+      setHasAllowance(true);
```
rafinskipg, line 41: `setHasAllowance(allowance.gte(depositAmount))` this makes it in one line

apps/webapp/src/modules/dsr/components/DsrWithdrawModal.tsx
rafinskipg, line 56: this is a bit crazy.. 🦗 

We should aim for tokenInput to accept directly whatever the hook returns, to avoid doing parse + format, we are doing 2 conversions there

apps/webapp/src/pages/Dsr.tsx
```diff
@@ -77,7 +78,7 @@ function Dsr(): React.ReactElement {
                           sources={dataSources}
                           message="The DSR APY is calculated by dividing the dsr by the seconds per year"
                         >
-                          <StatNumber>{formatUnits(data.dsrApy, 0)} (basis points)</StatNumber>
+                          <StatNumber>{formatPercent(data.dsrApy)}</StatNumber>
```
rafinskipg, line 81: 👍

packages/utils/src/math.test.ts
```diff
@@ -130,6 +130,26 @@ describe('Risk parameter math functions using ETH-A risk parameters', () => {
...
+    const expected = BigInt('1000105036786093739634792556');
+
+    expect(balance).eq(expected);
+  });
+
+  it('Can calculate the annual DAI savings rate', () => {
+    const rate = math.annualDaiSavingsRate(dsr);
+    const expected = BigInt('10000000622198524');
+
+    expect(rate).eq(expected);
```
rafinskipg, line 149: nice

rafinskipg (approved):

apps/webapp/src/modules/dsr/components/DsrDepositModal.tsx
b-pmcg: 🧠

PR #106, opened 2023-04-17
STart showing the staked dai in all farms

rafinskipg:
From the FE, it aggregates the data in all farms and calculates how much DAI the user has staked, and reflects it on the top summary.

TODO: 
- [ ] Update the amount of Subdaos the user staked to
- [ ] Refetch the balance after a stake and withdrawal
image


apps/webapp/src/modules/farms/components/StakedDaiSummary.tsx
```diff
@@ -9,9 +15,61 @@ export function StakedDaiSummary() {
...
+
+  const farms = useAvailableTokenFarms({
+    chainId
+  });
+
+  const provider = useProvider();
+
+  const [totalDaiStaked, setTotalDaiStaked] = useState(BigNumber.from(0));
+
+  useEffect(() => {
```
torhector2, line 40: Move this useEffect into a utils hook `useTotalDaiStaked`
It'll make the component easier to read and we can use it on other components (if needed too).

rafinskipg: yeah i was thinking the same, i was waiting to finish the implementation to see if it fits the design, and the move it to hooks if it makes sense

PR #107, opened 2023-04-18
Search farms

torhector2:
Allows to search farms by token name, symbol, token address, and farming address

Adds simple Farm Detail page and route

### SearchBar results
image
image
image

### Basic Farm Detail Page
image


apps/webapp/src/modules/search/components/SearchResultsListItem.tsx
```diff
@@ -55,7 +56,7 @@ function processSubtitle(subtitle: string | string[], maxTags = 3) {
 }
 
 export function SearchResultsListItem({ result, onClick }: Props) {
-  const image = useIpfsToHttp(result.image);
+  const image = isImageRoute(result.image) ? result.image : useIpfsToHttp(result.image);
```
rafinskipg, line 59: maybe the images should always come as image, so we don't have to do this here

PR #108, opened 2023-04-18
Search accounts

torhector2:
Allow to search by address accounts

- Prevent to show duplicates if the address is a contract related with farms
- Differentiate between user's address and other addresses
- They need to be a valid addresses

### Any address
image

### User address
image

### Also user address
image

### Invalid address
image

### Farm addresses don't appear duplicated
image


rafinskipg (approved):
good job

PR #109, opened 2023-04-18
Start splitting farm info in farm detail page

rafinskipg:
- Clicking any farm card will navigate to the page detail
- Split the "farm summary" that displays staked amounts and "manage position" into several components, 
- Added the not found component in case the farm is missing, 
- FarmDetail component

image
image


apps/webapp/src/modules/farms/components/FarmList.tsx
```diff
@@ -35,7 +36,11 @@ export function FarmList({ address }: { address?: `0x${string}` }): React.ReactE
       ) : (
         <Flex>
           {farms.map((farm, index) => (
-            <FarmItem key={index} farm={farm} ownerAddress={address} />
+            <Box key={`farm-${index}`}>
+              <Link as={ReactRouterLink} to={`/farms/${farm.contractAddress}/${farm.stakeToken.address}`}>
```
torhector2, line 40: the whole card acts as a link, I think we should define only some parts to do it, like the header and the icon. What do you think? also the button event is being captured by the link.

image

torhector2 (approved):
Looks very good 💯 

I leave my approved review ✅ so it's not a blocker, but take a look at my comment.

apps/webapp/src/modules/farms/components/FarmList.tsx
rafinskipg: yeah, i think we have to talk with tiago, the link is just temporary, in his designs is the button that acts as a link, but in general nowadays cards tend to be clickable too

PR #112, opened 2023-04-19
Subdao summary

rafinskipg:
image


apps/webapp/src/modules/farms/types/farm.d.ts
```diff
@@ -5,4 +5,10 @@ export type Farm = {
   rewardToken: Token;
   contractAddress: `0x${string}` | string;
   chainId: number;
+  subdao: {
```
torhector2, line 8: why don't we export the Farm type from the hooks package (as it's being defined there too) and imported in the webapp where needed from there?

torhector2 (approved):
Nice.

I left a comment, take a 👀

apps/webapp/src/modules/farms/types/farm.d.ts
rafinskipg: the current differnce is that the Token entity is different, on the hooks package the token is a string, here is a Token object that has more info,

I really want to move the Token definition to the hooks package to return full tokens probably. Maybe is time to do it

PR #113, opened 2023-04-19
Refactor tokens

rafinskipg:
We had this Token entity on the webapp that was pretty useful to access token information when you need something extra in addition to the token address.

Meanwhile we were developing the farms we found out that it was useful to hydrate the farms response with the full token info, but at the end a lot of code was spreading through the webapp to make this possible. 

It is easier if the token entity comes from the contracts package, since is tied to it's versioning and implementation. And then the hooks package can return the information already hydrated, simplifying the webapp source code overall.

👍 

```
Token = {
 address: string,
 name: string,
 symbol: string
 }
```


packages/hooks/src/vaults/useCollateralTypes.tsx
```diff
@@ -10,17 +11,17 @@ export function useCollateralTypes(): MakerReadHook & {
       {
         id: SupportedCollateralTypes.ETH_A,
         name: 'ETH-A',
-        symbol: 'ETH'
+        token: TOKENS[1].eth
```
rafinskipg, line 14: Reminder that this data structure is still temporary since we paused vaults

packages/contracts/src/index.ts
```diff
@@ -14,3 +21,86 @@ export enum ORACLE_CONTRACT_NAMES {
...
+};
+
+export const TOKENS: TokenMapping = {
+  [mainnet.id]: {
+    eth: {
+      address: ZERO_ADDRESS,
+      chainId: mainnet.id,
+      name: 'Ether',
+      symbol: 'ETH'
+    },
```
b-pmcg, line 38: what do you think about adopting the same pattern as the addresses in the wagmi generated file? So like
```
eth: {
    address: {
      [mainnet.id]: '0x123',
      [goerli.id]: '0x456'
    },
    name: 'Ether',
    symbol: 'ETH'
  }
```

apps/webapp/src/modules/farms/components/FarmList.tsx
```diff
@@ -1,15 +1,14 @@
...
 import LoadingPlaceholder from '../../ui/components/LoadingPlaceholder';
 import { FarmItem } from './FarmItem';
-import { useHydratedFarms } from '../hooks/useHydratedFarms';
 import { useNavigate } from 'react-router-dom';
 
 export function FarmList({ address }: { address?: `0x${string}` }): React.ReactElement {
   const chainId = useChainId();
 
-  const farms = useHydratedFarms(chainId);
+  const farms = useAvailableTokenFarms(chainId);
```
b-pmcg, line 11: Maybe we can move the `useChainId()` hook inside the `useAvailableTokenFarms()` to simplify the usage.

tyler17 (approved):

packages/contracts/src/index.ts
rafinskipg: good idea

apps/webapp/src/modules/farms/components/FarmList.tsx
b-pmcg: yeah that's true, in case for any reason they want to call the hook on a different chain than the one that is connected

PR #114, opened 2023-04-20
add svg in border to token

rafinskipg:
image


apps/webapp/src/modules/farms/components/StakedTokenIcons.tsx
```diff
@@ -0,0 +1,39 @@
+import { Box } from '@chakra-ui/react';
+import { Token } from '@makerdao-dux/contracts';
+import TokenIcon from '../../tokens/component/TokenIcon';
+
+export function StakedTokenIcons({ big, small }: { big: Token; small: Token }) {
```
torhector2, line 5: From the design I see some tokens have and some others don't have the svg border. Maybe we can add a `highlight` boolean prop to this component to show or not the coloured border.

torhector2 (approved):
LGTM

I left a comment that shouldn't block the merge of this PR, we can talk with Tiago to define that behaviour.

apps/webapp/src/modules/farms/components/StakedTokenIcons.tsx
rafinskipg: oki, as the design progresses and we move to highlighter farms and normal farms we can add this, thanks

PR #116, opened 2023-04-21
add hooks for total staked and calculate APR for farm

b-pmcg:
### What does this PR do?
- hook to get the total staked (`totalSupply`) from a farm
- hook to get the APR of a farm
- adds math functions to calculate APR using `rewardRate` and `totalSupply`

Screen Shot 2023-04-25 at 3 44 43 PM


packages/hooks/src/staking/useStakingRewardsApr.tsx
```diff
@@ -0,0 +1,84 @@
...
+import { ethers } from 'ethers';
+import { TRUST_LEVELS } from '../constants';
+
+type UseStakingRewardsAprResponse = MakerReadHook & {
+  data: bigint | undefined;
+};
+
+// TODO fetch prices from oracle hooks
+const subdaoUSDVal = ethers.BigNumber.from(1);
+const daiUSDVal = ethers.BigNumber.from(1);
```
b-pmcg, line 16: Without oracles contracts we have to hardcode these values. Not sure if this is the best place for it or not, but ultimately we will need to replace with the farms PIP contracts.

tyler17 (approved):

packages/hooks/src/staking/useStakingRewardsApr.tsx
```diff
@@ -0,0 +1,84 @@
...
+  } = useContractRead({
+    chainId: chainId as any,
+    address: contractAddress,
+    abi: stakingRewardsDaiABI,
+    functionName: 'rewardRate',
+    enabled: !!totalStaked
+  });
+
+  // Hook common interface
+  const isLoading = useMemo(() => {
```
torhector2, line 58: no need to use `useMemo` with booleans

torhector2 (approved):
💯 LGTM

Just one minor comment.

apps/webapp/src/modules/farms/components/FarmItem.tsx
```diff
@@ -15,20 +22,34 @@ export function FarmItem({
...
 
+  const { data: totalStaked } = useStakingRewardsTotalStaked({
+    chainId,
+    address: ownerAddress,
+    contractAddress: farm.contractAddress as `0x${string}`
+  });
+
+  const { data: apr } = useStakingRewardsApr({
+    chainId,
+    address: ownerAddress,
```
adamgoth, line 40: A wallet connection (owner address here) is required in order to show an APR?

apps/webapp/src/modules/farms/components/FarmMetrics.tsx
```diff
@@ -1,16 +1,26 @@
...
+}: {
+  farm: Farm;
+  totalStaked?: BigNumber;
+  apr?: bigint;
+}) {
   return (
     <Flex>
       <Box m={4}>
-        <Text>553.2</Text>
+        {!!totalStaked && <Text>{formatAmount(totalStaked.toBigInt())}</Text>}
```
adamgoth, line 18: Should we have a loading skeleton or some placeholder if `totalStaked` doesn't exist?

packages/hooks/src/staking/useStakingRewardsApr.tsx
```diff
@@ -0,0 +1,84 @@
...
+// TODO fetch prices from oracle hooks
+const subdaoUSDVal = ethers.BigNumber.from(1);
+const daiUSDVal = ethers.BigNumber.from(1);
+
+export function useStakingRewardsApr({
+  contractAddress,
+  chainId
+}: {
+  contractAddress: Address;
+  address?: Address;
```
adamgoth, line 23: Is this address value used somewhere? I don't see it being used but I could be overlooking it

adamgoth (commented):
Minor thing but looks weird to see such a large number include decimals. Maybe we come up with a helper for formatting numbers.

Something as simple as
less than 100 = 2 decimals // ex `10.34% APR`
greater than 100 = 0 deicmals // ex `134% APR`, `2,899% APR`

Screenshot 2023-04-26 at 12 45 45 PM

packages/utils/src/math.test.ts
```diff
@@ -150,6 +150,24 @@ describe('DSR Calculations', () => {
...
+  const rewardRate = BigInt('1157407407407407'); // .01 token per second
+  const totalStaked = BigInt('1000000000000000000000'); // 1000 tokens staked
+
+  it('Can calculate the APR', () => {
+    const rewardsRateValue = math.tokenValue(rewardRate, rewardsTokenPrice);
+    const totalStakedValue = math.tokenValue(totalStaked, stakingTokenPrice);
+
+    const apr = math.stakingRewardsApr(rewardsRateValue, totalStakedValue);
+    const expected = BigInt('36499999999999987152');
+
```
rafinskipg, line 166: why are the numbers so big? What is the representation? 

Shouldn't it be something easier like "30" for 30%? 

i always get confused with this things with the big decimals.

packages/utils/src/math.ts
```diff
@@ -131,9 +131,53 @@ export const annualDaiSavingsRate = (dsr: bigint): bigint => {
...
+  const yrvFixed = FixedNumber.fromValue(yearlyRewardsValue, WAD_PRECISION);
+  const tsvFixed = FixedNumber.fromValue(totalStakedValue, WAD_PRECISION);
+
+  const apr = yrvFixed.div(tsvFixed);
+
+  return apr.value;
+};
+
+// Calculate APY
+export const calculateApy = (apr: bigint): bigint => {
```
rafinskipg, line 164: I added this, but i'm not sure this is correct, and it overflows when passing those bigints

rafinskipg (changes requested):
I'm unsure about the output of the APR i would like to understand it better

b-pmcg:
> Minor thing but looks weird to see such a large number include decimals. Maybe we come up with a helper for formatting numbers.
> 
> Something as simple as less than 100 = 2 decimals // ex `10.34% APR` greater than 100 = 0 deicmals // ex `134% APR`, `2,899% APR`
> 
> Screenshot 2023-04-26 at 12 45 45 PM

yeah, this makes sense. I started working on it a little bit, but it's actually pretty complicated, so I think it should be a separate ticket since it affects all numbers. The reason it's complicated is because `parseFloat` returns a number with an exponent in it if there are too many decimal places (eg `1.44e-11`) and doing gt/lt comparisons on this fails.

We can probably develop a better formatter that has a set of rules that the numbers filter through to get to the results we want.

b-pmcg:
accidentally click closed & comment 🤦

packages/utils/src/math.test.ts
b-pmcg: answered in the discord as well, but generally its because the values we get directly from the contract are bigints with 18 decimal places and I wanted to keep the math functions pure so they always deal with the values from the contracts. And the UI is able to convert the bigints to more human readable forms pretty easily.

packages/utils/src/math.ts
b-pmcg: I think your math is correct, but I had the same issue with the overflows. I think using APR makes the most sense with our farms due to the reasons you mentioned.

packages/utils/src/math.test.ts
rafinskipg: so it doesn't matter if you send a numer with 18 decimals or just any number right? it will calculate the relative APRs correctly?

rafinskipg (approved):

PR #119, opened 2023-04-24
force open localhost on bundler

rafinskipg:
When launching bundle, autoopen localhost:8080 on the browser


tyler17:
In general do I need to do something to launch the bundle after running `pnpm bundle`?

rafinskipg:
> In general do I need to do something to launch the bundle after running `pnpm bundle`?

you have to go to the bundler-webapp folder, and theres a bundle folder with the binaries, 

you can run them from your shell

tyler17:
Ah right, thanks.  Works for me on my mac

tyler17 (approved):

PR #120, opened 2023-04-24
Theme customization

adamgoth:
### What does this PR do?

* Consolidate list of current color tokens and give things more generic names
* Example: `maker` becomes `accentOne`, `backgroundGray` becomes `secondaryBackground`
* Fix disabled button prop name

Merged with Hector's PR (notes below):
> Not quite sure about the approach so I'd appreciate feedback.
> Not sure about the naming of the different variants either.
> I started by allowing cutomize the color, but we can add the font size too maybe.
> Would you create different variants based on these h1, h2, ... body large and the main colors? Or we use a combination of a variant and sx/color in a component? I like this one more: `<Text variant="medium" color="maker">`

### TODOs:
- [x] Define full list of colors -- use `primary`, `onPrimary`, `primaryEmphasis` pattern, `secondary`, `onSecondary` etc
    current list (still work in progress):
    primary
    onPrimary
    primaryEmphasis
    secondary
    onSecondary
    secondaryEmphasis
    text
    secondaryText
    background
    secondaryBackground
    borderColor
    accentOne
    accentTwo
    success
    error
    warning
    info
- [x] Keep font out of config for now, we will figure out an approach for this separately 
- [x] Add typography components to theme page
- [x] Add config option for wallet connect button and text colors (similar to `textColor`, `linkColor`)
- [x] Fix rainbow button colors to use config colors (move to app code)
- [x] Remove references to specific colors in app (ex: `white`, `yellow.500`, `colorScheme={}`) and replace with defined color tokens 
- [x] Update components in app to use new typography variants
- [x] Update toasts with success, error, warning, info colors
- [x] Update alerts with success, error, warning, info colors


rafinskipg (approved):
nice, just the comment about using maybe storybook (that we are not mantaining really.. )

apps/webapp/src/modules/ui/headingTheme.ts
```diff
@@ -1,13 +1,38 @@
...
+  color: defaultConfig.headingColor,
+  fontSize: '24px',
+  fontWeight: 'regular',
+  lineHeight: '36px',
+  letterSpacing: 0
 });
 
 export const headingTheme = defineStyleConfig({
-  variants: { h1 }
+  variants: { xl: h1, lg: h2, md: h3, sm: h4 }
```
b-pmcg, line 37: nice, is it possible to add a default value? So if the most commonly used size is `md` we can just do `<Heading>...` without explicitly setting the variant.

b-pmcg: just noticed the default-config file, maybe that's a better place

tyler17:
looking good 👍

apps/webapp/src/modules/account/components/AccountDetail.tsx
```diff
@@ -57,7 +57,7 @@ export function AccountDetail({ address }: { address: `0x${string}` }): React.Re
       </Text>
       <Card my={2}>
         <CardHeader>
-          <Heading size="lg">DS Proxy</Heading>
+          <Heading>DS Proxy</Heading>
```
b-pmcg, line 60: ❤️

b-pmcg (approved):
left a minor comment but otherwise looks great

apps/webapp/src/modules/config/components/IPFSGatewaySection.tsx
```diff
@@ -11,7 +11,7 @@ export function IPFSGatewaySection({
   return (
     <Card>
       <CardHeader>
-        <Heading size="lg">IPFS Gateway</Heading>
+        <Heading sx={{ fontWeight: 'bold' }}>IPFS Gateway</Heading>
```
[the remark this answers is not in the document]
torhector2: Adam and me talked about this before, we decided to go for variants without fontWeight as the design system have multiple variants: h1 regular, h1 bold, h2 regular, h2 bold...

I guess we can talk about having "lg", "lg-bold", "md", "md-bold"... we can decide that with the other devs.

PR #121, opened 2023-04-24
Text style customization

torhector2:
Not quite sure about the approach so I'd appreciate feedback.
Not sure about the naming of the different variants either.

I started by allowing cutomize the color, but we can add the font size too maybe.

Would you create different variants based on these h1, h2, ... body large and the main colors? Or we use a combination of a variant and sx/color in a component? I like this one more:
`<Text variant="medium" color="maker">`

image


apps/webapp/src/modules/ui/headingTheme.ts
```diff
@@ -1,13 +1,51 @@
...
+  letterSpacing: 0
+});
+const h4Bold = defineStyle({
+  ...h4,
+  fontWeight: 'bold'
 });
 
 export const headingTheme = defineStyleConfig({
-  variants: { h1 }
+  variants: { h1, h1Bold, h2, h2Bold, h3, h3Bold, h4, h4Bold }
```
adamgoth, line 50: Maybe we should give these names with sizes, similar to text and links. h1 = xl, h2 = lg, h3 = sm, h4 = xs. Wdyt?

apps/webapp/src/modules/ui/headingTheme.ts
```diff
@@ -1,13 +1,51 @@
...
+  fontWeight: 'regular',
   lineHeight: '62px',
-  fontWeight: 700,
-  fontSize: '3rem'
+  letterSpacing: '-1px'
+});
+const h3Bold = defineStyle({
+  ...h3,
+  fontWeight: 'bold'
+});
```
adamgoth, line 36: Sort of on the fence about whether we should create an entire new variant just to make it bold, of if we should just add that directly when needed. Wdyt?

apps/webapp/src/modules/ui/linkTheme.ts
```diff
@@ -20,5 +41,5 @@ const nostyle = defineStyle({
 });
 
 export const linkTheme = defineStyleConfig({
-  variants: { primary, nostyle }
+  variants: { large, medium, mediumBold, small, nostyle }
```
adamgoth, line 44: Not super important but I think we should stick with how chakra names sizes, `sm`, `md`, `lg`, `xl`, `2xl`, etc. Just looks cleaner and easier to type 😄

apps/webapp/src/modules/config/default-config.ts
```diff
@@ -21,6 +21,14 @@ export const defaultConfig: SiteConfig = {
     cardBorder: { default: 'white', _dark: '#434343' },
     maker: { default: '#1aab9b' }
   },
+  fonts: {
+    heading: '"Plus Jakarta Sans", "sans-serif"',
+    body: '"Plus Jakarta Sans", "sans-serif"'
+  },
```
adamgoth, line 27: I was trying to think about the best way to handle this, because this only works since we have `@fontsource/plus-jakarta-sans` installed as a package. Maybe we just need to explain in the README about how to add custom fonts.

apps/webapp/src/modules/config/default-config.ts
```diff
@@ -21,6 +21,14 @@ export const defaultConfig: SiteConfig = {
...
+  headingColor: 'text',
+  formLabelColor: 'text',
+  linkColor: 'text',
+  textColor: 'secondaryText',
```
adamgoth, line 31: I think this is a good idea, allows some extra customization.

Shouldn't `textColor` be `text` instead of `secondaryText` though? I thought that was the common text color.

adamgoth (commented):
Nice work so far 👍  Left a few thoughts. Let me know what you think.

apps/webapp/src/modules/ui/headingTheme.ts
torhector2: I like it but also it'll make harder to know which variant to apply based on Figma (where they're referenced as h1 h1 bold, h2...)

apps/webapp/src/modules/ui/headingTheme.ts
torhector2: yes I had my doubts here as well, maybe the same as with colors, we can apply the bold style directly on the component when needed.

apps/webapp/src/modules/config/default-config.ts
torhector2: I had the impression that on Figma most of body text had secondaryText color. Weird I know. I should confirm with Tiago.

apps/webapp/src/modules/config/default-config.ts
torhector2: true, this isn't as easy as adding it to the config file as they need to install the font as you said, but also import them in a couple of files `main.tsx` for the `rainbowThemeConfig` and in the **storybook** `preview.jsx `(which I'm not sure they'll use that much).

apps/webapp/src/modules/ui/headingTheme.ts
```diff
@@ -1,13 +1,38 @@
...
+  fontSize: '24px',
+  fontWeight: 'regular',
+  lineHeight: '36px',
+  letterSpacing: 0
 });
 
 export const headingTheme = defineStyleConfig({
-  variants: { h1 }
+  variants: { xl: h1, lg: h2, md: h3, sm: h4 }
 });
```
adamgoth, line 38: add a default of `md` maybe?

adamgoth:
Merged with https://github.com/makerdao-dux/metadao-frontend/pull/120

PR #124, opened 2023-04-27
WIP update wagmi to v1

b-pmcg:
### What does this PR do?
- upgrade wagmi to v1
- migrate ethers provider pattern to wagmi client pattern
- replace all ethers functions with viem equivalents
- update all BigNumber instances to bigint
- update hooks to handle viem contract responses

### Steps for testing:
- run test suite
- manual QA inputs like DSR deposit, farming stake


package.json
```diff
@@ -7,7 +7,6 @@
   },
   "scripts": {
     "preinstall": "npx only-allow pnpm",
-    "postinstall": "pnpm --filter './packages/*' run build",
```
b-pmcg, line 10: Temporary, had to remove this to install the packages, remember to re-add

packages/hooks/src/dsProxy/useDsProxyData.ts
```diff
@@ -20,7 +20,7 @@ export function useDsProxyData(owner?: `0x${string}`): DsProxyHookResponse {
 
   const acct = owner || connectedAddress || ZERO_ADDRESS;
 
-  const { address: contractAddress } = useProxyRegistry({
+  const { address: contractAddress } = getProxyRegistry({
```
torhector2, line 23: I wonder if these calls to getFoo should be wrapped in a useCallback, or useMemo, or write our own useContract hook that accepts the function and args to memoized the results.

packages/utils/src/formatValue.ts
```diff
@@ -20,13 +20,13 @@ const formatUnitPercent = new Intl.NumberFormat('en-US', {
   maximumFractionDigits: 0
 });
 
-export function formatAmount(amount: bigint, unit: 'wad' | 'ray' = 'wad') {
+export function formatAmount(amount: bigint, unit: 'wad' | 'ray' = 'wad'): `${number}` {
   const amountToFormat = unit === 'ray' ? formatRay(amount) : formatWad(amount);
 
-  return formatUnitAmount.format(parseFloat(amountToFormat));
+  return formatUnitAmount.format(parseFloat(amountToFormat)) as `${number}`;
 }
```
b-pmcg, line 27: This return type is because `` `${number}` `` is what vite `parseUnits()` functions expect

packages/hooks/src/dsProxy/useDsProxyData.ts
b-pmcg: it turns out the viem Contract instance no longer has the address prop but we can now import the static addresses directly from the generated file.

packages/hooks/src/vaults/useCollateralData.ts
```diff
@@ -43,22 +42,19 @@ export function useCollateralData(
     chainId: chainId as any,
     functionName: 'ilks',
     args: [ilk],
-    scopeKey: utils.parseBytes32String(ilk),
-    enabled: !!ilk && ilk !== ZERO_BYTES32,
-    select(data) {
-      // Append the ilk to the response so we can match it later
-      return { ...data, ilk };
-    }
```
b-pmcg, line 51: this wasn't really necessary

packages/hooks/src/vaults/useCollateralData.ts
```diff
@@ -43,22 +42,19 @@ export function useCollateralData(
...
+    scopeKey: hexToString(ilk, { size: 32 }),
+    enabled: !!ilk && ilk !== ZERO_BYTES32
   });
 
+  const [ilkArt, rate, spot, line, dust] = vatIlkData || [];
```
b-pmcg, line 49: viem contract responses don't give us named variables like ethers did, so we can no longer dot off `vatIlkData` as before.

jxom: There is an explanation for this FYI: https://github.com/wagmi-dev/viem/blob/main/.github/FAQ.md#why-is-a-contract-function-return-type-returning-an-array-instead-of-an-object

adamgoth (commented):
Lots of updates! Nice job 🔥 

Looks pretty solid to me. I found a few minor issues (some not related to your code) and started a PR into this one here: https://github.com/makerdao-dux/metadao-frontend/pull/148

package.json
```diff
@@ -69,5 +71,10 @@
     "npm-run-all": "^4.1.5",
     "path": "^0.12.7",
     "prettier": "^2.8.1"
+  },
+  "pnpm": {
+    "overrides": {
+      "nwsapi": "2.2.2"
+    }
```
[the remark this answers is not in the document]
torhector2: it forces a dependency to have a particular version independently of what version a library uses.

rafinskipg (approved):

rafinskipg (approved):

adamgoth (commented):
In testing, I've had two transactions run out of gas while trying to approve a custom spend limit of 25 DAI for depositing. Have you had this happen to you in testing?

https://goerli.etherscan.io/tx/0xeb12f76c1e35a6a8f2cb39176daff21b57f0980145cd54d4a33760794467c8b2

adamgoth (commented):
One last thought -- it could be helpful to add a section to the README about our approach with math and formatting with numbers (`BigInt` vs `BigNumber` etc) since we're using BigInt everywhere now including with component state.

packages/hooks/src/decentralizedStorage/useEnsContent.ts
```diff
@@ -31,7 +31,7 @@ export function useEnsContent({
   const { error, data, isValidating } = useSWR(
     resolver ? [chainId, `${subdomain}.${name}`] : null,
     async () => {
-      const val = await resolver?.getContentHash();
+      const val = await resolver;
```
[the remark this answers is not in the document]
b-pmcg: Because the type of `resolver` is a `0x${string}`, which I believe is the content hash itself, but should probably verify that.

packages/hooks/src/vaults/useCollateralData.ts
b-pmcg: good to know, thank you for providing that context @jxom !

packages/hooks/src/dsr/useDsrData.ts
```diff
@@ -120,7 +117,14 @@ export function useDsrData(address?: `0x${string}`): DsrHook {
...
   const data = useMemo<DsrHookData | undefined>(() => {
-    if (!pie || !Pie || !dsr || !chi || !userDaiBalance) {
+    // We need to assert the balances below are not 0n or we'll get stuck loading
+    if (
+      (!pie && pie !== 0n) ||
+      !Pie ||
+      (!dsr && dsr !== 0n) ||
+      !chi ||
+      (!userDaiBalance && userDaiBalance !== 0n)
+    ) {
```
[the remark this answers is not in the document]
b-pmcg: realistically it would mean the dsr is set to 0 and there is 0 dai joined. Unlikely but we should probably cover that case since its just 2 lines of code

package.json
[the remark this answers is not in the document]
b-pmcg: there is an issue with chakra and react testing library that was caused by a newer version of the dependency `nwsapi`, issue for reference: https://github.com/dperini/nwsapi/issues/81

PR #125, opened 2023-04-28
add balance error to TokenInput

tyler17:
### What does this PR do?
-Adds a `isBalanceError` and `tokenName` field to the TokenInput component, which are used to display a balance error message
-disable button if balance error


apps/webapp/src/modules/ui/components/TokenInput.tsx
```diff
@@ -9,14 +9,18 @@ export function TokenInput({
...
+  isBalanceError,
+  tokenName = ''
 }: {
   value: BigNumber;
   onChange: (val: BigNumber) => void;
   balance: BigNumber;
   placeHolder?: string;
   title?: string;
   dataTestId?: string;
+  isBalanceError: boolean;
```
torhector2, line 22: I think we could change from isBalanceError to simple accepting an error object. Then print it's message in the FormErrorMessage.
`<FormErrorMessage>{error.message}</FormErrorMessage>`

This way we can be more dynamic about which error show to the user.

torhector2 (commented):
Looking good, just one comment.

apps/webapp/src/modules/ui/components/TokenInput.tsx
tyler17: Yeah probably makes sense, I updated it 👍

PR #127, opened 2023-04-28
use subgraph farms

rafinskipg:
Uses the subgraph at 
 https://thegraph.com/hosted-service/subgraph/makerdao-dux/subdao-farms

https://github.com/makerdao-dux/subdao-farms 

Displays the real amount of depositors

image


torhector2 (approved):
🎸

apps/webapp/src/modules/farms/components/FarmMetrics.tsx
```diff
@@ -37,9 +39,14 @@ export function FarmMetrics({ farm }: { farm: Farm }) {
           <Text variant="sm">TVL</Text>
         </Box>
         <Box m={4}>
-          <Text variant="md" color="text">
-            3
-          </Text>
+          {!farmHistory ? (
```
adamgoth, line 42: Looks like the hook returns a loading prop we can use. We should handle an error coming back from the hook too.

packages/hooks/src/staking/history/useFarmHistory.tsx
```diff
@@ -0,0 +1,85 @@
+import { request, gql } from 'graphql-request';
+import useSWR from 'swr';
+import { FarmHistory, FarmHistoryRaw } from './farmHistory';
+import { goerli } from 'wagmi';
+
+const urlSubgraphGoerli = 'https://api.thegraph.com/subgraphs/name/makerdao-dux/subdao-farms';
+const urlSubgraphMainnet = 'https://api.thegraph.com/subgraphs/name/makerdao-dux/subdao-farms-mainnet';
```
adamgoth, line 7: Not super important but maybe add a `TODO` to replace this URL later

rafinskipg: will thanks

PR #131, opened 2023-05-02
Webapp icons pkg

torhector2:
Creates MakerDUX Chakra Icons package.

We separate the icons from the webapp creating a new package users can use in any app.

The package is defined as `"sideEffects": false` so Vite can perform "tree shaking" during the build process and remove all unused icons.


apps/webapp/src/modules/config/default-config.ts
```diff
@@ -47,6 +47,7 @@ export const defaultConfig: SiteConfig = {
   linkHoverColor: 'primaryEmphasis',
   linkActiveColor: 'primaryEmphasis',
   textColor: 'secondaryText',
+  iconColor: 'primary',
```
adamgoth, line 50: 👍

packages/chakra-icons/package.json
```diff
@@ -0,0 +1,56 @@
...
+    "@types/react-dom": "^18.0.3",
+    "jsdom": "^21.0.0",
+    "react": "^18.1.0",
+    "react-dom": "^18.1.0",
+    "typescript": "^4.9.3",
+    "vite-plugin-dts": "^1.7.1",
+    "@chakra-ui/anatomy": "^2.1.1",
+    "@chakra-ui/icons": "^2.0.16",
+    "@chakra-ui/react": "^2.6.0",
+    "@testing-library/react": "^13.3.0"
```
adamgoth, line 46: Is this dep being used? Don't think the package has tests

adamgoth (approved):
Looks good to me. Thanks for taking care of this. Only thing I'm wondering is if the name of the package should be `chakra-icons`, or just `icons`, because the custom icons aren't coming from chakra. Does the package work in a React app that is not using Chakra? If so, I think we should remove `chakra` from the package name. Thoughts?

packages/chakra-icons/package.json
torhector2: I added one to test one icon is being rendered. I can remove it though, it's not testing anything else.

torhector2:
All icons have chakra as a dependency.

import { IconProps, Icon } from '@chakra-ui/react';

But maybe we can export the icons wrapped in chakra Icon components but also export an object with only the path so people can use it in other compatible UI library (like MUI, that has SVGIcon and can wrap a path)

{
  AlertOctagon: {
    path: ...
  },
  AlertSquare: {...}
}

rafinskipg (approved):
neat, ain't it?

PR #135, opened 2023-05-04
Decrease useprepare hooks

b-pmcg:
### What does this PR do?
- allows passing `enabled` prop into hooks so consumers can pass their own logic
- updates various usePrepare hooks to be disabled when not necessary to run
- sets `staleTime` for read hooks to decrease background fetching


apps/webapp/src/modules/dsr/components/DsrDepositModal.tsx
```diff
@@ -39,8 +41,11 @@ export function DsrDepositModal() {
 
   const hasProxy = !!dsProxy?.dsProxyAddress;
 
+  const readyToDeposit = hasProxy && hasAllowance;
+
   const dsrDeposit = useDsrDeposit({
```
rafinskipg, line 46: I wonder if we could just put the enabled flag inside the useDSRDeposit hook, we could make the useDSRDeposit check if it has allowance and if it has proxy internally. since it will always need a proxy and allowance,

apps/webapp/src/modules/dsr/components/DsrDepositModal.tsx
```diff
@@ -101,6 +106,7 @@ export function DsrDepositModal() {
   });
 
   const dsProxyBuild = useDsProxyBuild({
+    enabled: !hasProxy,
```
rafinskipg, line 109: same as before, we could have the check inside the build proxy hook

apps/webapp/src/modules/dsr/components/DsrWithdrawModal.tsx
```diff
@@ -2,21 +2,31 @@ import { useState } from 'react';
...
+
+  const hasProxy = !!dsProxy?.dsProxyAddress;
+  const hasDsrBalance = dsrData && dsrData?.userDsrBalance > 0n;
+
+  const readyToWithdraw = hasProxy && hasDsrBalance;
+
   const dsrWithdraw = useDsrWithdraw({
-    amount: withdrawAmount,
+    amount: debouncedWithdrawAmount,
+    enabled: readyToWithdraw,
```
rafinskipg, line 29: same case

rafinskipg (changes requested):
me likey, but see a possible improvement

apps/webapp/src/modules/dsr/components/DsrDepositModal.tsx
torhector2: But if we remove the enabled flag from these hooks we will block consumers from using it, maybe they want to apply new and different constraints to consider the hook enabled.

I agree though that maybe we can move inside that `const readyToDeposit = hasProxy && hasAllowance;` and leave the user pass whatever they need to the enabled param.
So inside `useDSRDeposit` in particular (but it would be the same for other like this one) we would have a 

`enabled: !!enabled && !!dsProxy?.data?.dsProxyAddress && readyToDeposit,`

torhector2 (approved):
I like the enabled usage in the hooks 💯 🚀

I approve but please consider Rafa's comments.

apps/webapp/src/modules/dsr/components/DsrDepositModal.tsx
b-pmcg: good idea guys, I'll leave `enabled` as an optional boolean so consumers can use their own logic when necessary, but I agree in some cases the check can just be done inside the hook.

b-pmcg:
@rafinskipg @torhector2 I made the requested changes. Some tests are failing in the remote workflow but they're passing for me locally, and they seem to be related to vaults so idk if its affected by this PR.

rafinskipg (approved):

PR #137, opened 2023-05-05
Toast alert new design

torhector2:
I tried to use the alerts theme to do this customization, but our custom alert and toast designs have multiple layouts and components. That's why I decided to create new components for this.

### ToDo
- Think and define colors in the config file

### Custom Toasts
image
image

### Custom Alerts
image
image


b-pmcg:
looks really good! The only thing that looks a little strange to me that the shorter alerts have the icon above the content, while the longer ones have it in the left margin. I know that how it is on the designs, but maybe we can ask Tiago if that's what he intended, or see if everyone else thinks it's fine.

I'm not totally familiar enough with Chakra yet to understand why we couldn't use the default chakra alert component with some different variants, but if its the only way then 👍

apps/webapp/src/modules/ui/components/Alerts.tsx
```diff
@@ -0,0 +1,226 @@
...
+  );
+}
+
+interface UseCustomToastOptions extends UseToastOptions {
+  link?: CustomLink;
+}
+
+type CustomToastFn = (options?: UseCustomToastOptions | undefined) => ToastId;
+
+export function useToast(options?: UseToastOptions): CustomToastFn {
```
b-pmcg, line 224: maybe we should move this to `modules/ui/hooks/useToast.ts`

torhector2: Agree 👍

apps/webapp/src/modules/ui/components/Alerts.tsx
```diff
@@ -0,0 +1,174 @@
...
+export function CustomToast({ size = 'lg', ...props }: Props): React.ReactElement {
+  return (
+    <BaseComponent variant="maker-toast" size={size} width={size === 'lg' ? '557px' : '466px'} {...props} />
+  );
+}
+
+export function CustomAlert({ size = 'lg', ...props }: Props): React.ReactElement {
+  return (
+    <BaseComponent
+      variant="maker"
```
rafinskipg, line 167: i'm not sure we should keep naming things "maker"

apps/webapp/src/modules/ui/components/Alerts.tsx
```diff
@@ -0,0 +1,174 @@
...
+  width,
+  title,
+  description,
+  link,
+  isClosable,
+  size,
+  variant
+}: Props): React.ReactElement {
+  const showInNewLine = size === 'sm';
+  const { main, LeftIcon } = useAlertUIData(title, description, status === 'loading' ? 'info' : status, link);
```
rafinskipg, line 118: why is this a hook ?

apps/webapp/src/modules/ui/components/Alerts.tsx
```diff
@@ -0,0 +1,174 @@
...
+const getTitle = (title: React.ReactNode) =>
+  typeof title === 'string' ? (
+    <Text variant="noStyle" sx={{ fontSize: '16px', lineHeight: '26px', fontWeight: 700 }}>
+      {title}
+    </Text>
+  ) : (
+    title
+  );
+
+const getDescription = (description: React.ReactNode) =>
```
rafinskipg, line 52: why we need this as functions? It could probably be inline

apps/webapp/src/modules/ui/components/Alerts.tsx
```diff
@@ -0,0 +1,174 @@
...
+    const titleComponent = getTitle(title);
+    const descriptionComponent = getDescription(description);
+    const linkComponent = getLink(link, status);
+    const LeftIcon = StatusIcons[status || 'info'];
+    const main = (
+      <>
+        <AlertTitle>{titleComponent}</AlertTitle>
+        <AlertDescription>
+          {descriptionComponent}
+          {linkComponent}
```
rafinskipg, line 91: i think this "linkComponetn" could just be a children component,

the "learn more" style link with the external icon it should be probably a "external  link" component that is reused accross the app, with the little arrow

rafinskipg: Not all the toasts will have a link . Example: "invalid network, change network to continue"

apps/webapp/src/modules/ui/hooks/useToast.tsx
```diff
@@ -0,0 +1,15 @@
+import { UseToastOptions, useToast as useChakraToast, ToastId } from '@chakra-ui/react';
+import { CustomLink, CustomToast } from '../components/Alerts';
+
+interface UseCustomToastOptions extends UseToastOptions {
+  link?: CustomLink;
```
rafinskipg, line 5: i think accepting a children component maybe better than a "link" , but idk. what other devs think? Would it be weird?

apps/webapp/src/modules/config/default-config.ts
```diff
@@ -22,10 +22,14 @@ export const defaultConfig: SiteConfig = {
...
-    success: { default: '#68D391' },
-    error: { default: '#E53E3E' },
-    warning: { default: '#ECC94B' },
-    info: { default: '#BEE3F8' }
+    success: { default: '#039855' },
+    error: { default: '#DF2020' },
+    warning: { default: '#FFA64C' },
+    info: { default: '#846EF7' },
+    infoBackground: { default: '#FCFAFF' },
+    successBackground: { default: '#F6FEF9' },
```
rafinskipg, line 30: i think we will need to define the dark colors

apps/webapp/src/modules/ui/components/Alerts.tsx
```diff
@@ -0,0 +1,174 @@
...
+    title
+  );
+
+const getDescription = (description: React.ReactNode) =>
+  typeof description === 'string' ? <Text variant="novariant">{description}</Text> : description;
+
+const getLink = (link?: CustomLink, status: Status = 'info') => {
+  if (!link) return null;
+
+  const lighterColor = adjustColor(defaultConfig.colors[status as Status].default, 15);
```
rafinskipg, line 58: is there a way to not use javascript and use ` filter: brightness(85%);` ?  idk if that would work

rafinskipg (changes requested):
I'm not sure about some things, i'll wait for other devs input

apps/webapp/src/modules/ui/components/Alerts.tsx
torhector2: right, any idea? "brand"?

apps/webapp/src/modules/ui/components/Alerts.tsx
torhector2: true, they were more complex before moving stuff into the theme. I can inline them.

apps/webapp/src/modules/config/default-config.ts
torhector2: I'll ask for them 👍

apps/webapp/src/modules/ui/components/Alerts.tsx
torhector2: As Deniz suggested we keep "maker" until a new brand is defined.

apps/webapp/src/modules/ui/components/Alerts.tsx
```diff
@@ -0,0 +1,119 @@
...
+  AlertDescription,
+  VStack
+} from '@chakra-ui/react';
+import { AlertCircle, AlertTriangle, CheckCircle, InfoCircle, XClose } from '@makerdao-dux/icons';
+
+export type CustomLink = { url: string; text: string };
+
+export interface CustomToastProps extends ToastProps {
+  onClose?: () => void;
+  renderBottom?: (status: ToastProps['status']) => React.ReactNode;
```
torhector2, line 19: I add this so we can define dynamic components based on the status in case we only know the status at runtime.

Not sure if it's helpful or weird.

rafinskipg (approved):
I'll approve but i would wait for @adamgoth  and @b-pmcg  to chip in, since they are the most experimented with chackra ui

apps/webapp/src/modules/config/default-config.ts
adamgoth: There are a lot of other colors that still need dark definitions. I think we should just add a backlog ticket for it for Tiago to define and then for devs to implement.

adamgoth (approved):
lgtm, nice job -- I asked a question in our design channel to see if we can maybe remove all the "classic" toast/alert stuff

PR #141, opened 2023-05-11
Fix dsr balance representation

torhector2:
DSR Balances were being represented wrong.
I think it'll be a good idea if @b-pmcg takes a look at this as he know better all the math.

### Before
image

### Now
image


torhector2:
@b-pmcg I've updated this branch with `main` after merging the `components` package. Once you have the `wagmiv1` PR merged you can merge this one as you suggested.

Also I added back the `decimals` optional param to `TokenInput` so we can use it with other tokens like USDC that has 6 decimals. Also so other devs using the components package can use it with any token. Let me know wyt.

b-pmcg (approved):
@torhector2 I changed the target to the `upgrade-wagmi-v1` branch and I merged that branch into here so it's all up to date. I think it's ready to be merged now. If the wagmi branch gets merged into main first you can just reset the target back to `main`.

PR #144, opened 2023-05-15
convert ray unit to wad inside dsrBalance math fn

b-pmcg:
### What does this PR do?
updates the dsrBalance math fn to convert ray unit to wad similar to others


b-pmcg:
The only times `ray` units seem to be used is when calculating percentages like rates and fees. It has higher decimal places for more precision, but as @tyler17 mentioned in the notion doc, the user probably will never need to input a number at that level of precision since most tokens balances are 18 decimals, so I think its fair that the `TokenInput` component should not have to expect anything other than an 18 decimal `wad`, and so we can just normalize it.

apps/webapp/src/modules/dsr/components/DsrWithdrawModal.tsx
```diff
@@ -61,7 +61,7 @@ export function DsrWithdrawModal() {
       <Flex sx={{ flexDirection: 'column', maxWidth: 250 }}>
         <TokenInput
           title="Withdraw Amount"
-          balance={parseUnits(formatWad(dsrData?.userDsrBalance).toString(), 18) || BigNumber.from(0)}
+          balance={parseUnits(formatWad(dsrData?.userDsrBalance)) || BigNumber.from(0)}
```
b-pmcg, line 64: This ugliness will become unnecessary when we merge the wagmi v1 because `balance` prop will become a bigint and we can just pass `dsrData?.userDsrBalance` directly

apps/webapp/src/modules/ui/components/TokenInput.tsx
```diff
@@ -23,13 +22,13 @@ export function TokenInput({
...
   // The input value should be able to be changed by the user in any way, and only trigger the change when the units are correct.
-  const [inputValue, setInputValue] = useState(formatUnits(value, decimals));
+  const [inputValue, setInputValue] = useState(formatUnits(value));
 
   const [errorInvalidFormat, setErrorInvalidFormat] = useState(false);
 
   const getBigNumberFromInputString = (val: string): BigNumber => {
     const formattedVal = parseFloat(val.replaceAll(',', '.'));
-    const newValue = val ? parseUnits(formattedVal.toString(), decimals) : parseUnits('0', decimals);
+    const newValue = BigNumber.from(val ? parseUnits(formattedVal.toString()) : '0');
```
b-pmcg, line 31: This whole file will change a lot when we update wagmi, we won't use bignumber and won't need to use the built in math fns like mul/sub/add

torhector2 (approved):
lgtm, I tried the dsr and the farms and the formats look good.

My only question is if we want to extend TokenInput to accept any possible decimals or limit it to 18.

adamgoth (approved):

adamgoth (approved):

PR #152, opened 2023-05-23
DSR to SSR

adamgoth:
Updates the DSR to use the SSR contract. Renames things. Removes proxy contract logic from deposit/withdraw.


b-pmcg (approved):

b-pmcg:
nice that we don't have to worry about setting up a proxy for SSR

packages/hooks/src/ssr/useSsrWithdraw.ts
```diff
@@ -1,62 +1,36 @@
...
-  // Only enabled if user has a balance in the DSR and they have a proxy
-  const enabled = dsrData && dsrData?.userDsrBalance > 0n && !!dsProxy?.data?.dsProxyAddress;
-
-  // encode withdraw function call for dsProxy to execute
-  const encodedJoinData = setMax
-    ? encodeFunctionData({
-        functionName: 'exitAll',
-        abi: proxyActionsDsrABI,
-        args: [mcdJoinDaiAddress[chainId], mcdPotAddress[chainId]]
-      })
```
b-pmcg, line 43: I think the way to implement a similar function to `exitAll` would be something like call [maxWithdraw](https://github.com/hexonaut/savings-dai/blob/master/src/SavingsDai.sol#L317) to calculate the exact amount and pass that value to the withdraw function.

rafinskipg: yeah , +1 to what @b-pmcg  said here, we could keep the maxWithdraw

rafinskipg (approved):

packages/hooks/src/ssr/useSsrWithdraw.ts
adamgoth: I actually had that in there initially, but it seemed to work just fine without it. It also requires more updates to the UI and token input to indicate max withdraw. We had the `setMax` option in the hooks, but the UI was never sending `setMax` as `true` to the hook.

I'll add a TODO and we add a separate story for it.

PR #153, opened 2023-05-25
Farm overview card

tyler17:
### What does this PR do?
Update farm overview card
before
Screen Shot 2023-05-24 at 9 35 50 PM
after
Screen Shot 2023-05-24 at 9 34 15 PM

### Link to Shortcut ticket:
https://app.shortcut.com/dux-makerdao/story/2124/polish-the-ui-on-the-farm-detail-page


tyler17:
Hm, on the vercel deployment the page doesn't load and there's this error: "Error: No wagmi client found."

apps/webapp/src/modules/farms/components/FarmItem.tsx
```diff
@@ -35,11 +35,18 @@ export function FarmItem({
...
+  });
+
+  // Rewards balance
+  const { data: rewardsBalance } = useStakingRewardsRewardsBalance({
+    chainId,
+    address: address,
+    contractAddress: farm.contractAddress as `0x${string}`
   });
 
   const isOwner = address === ownerAddress;
```
tyler17, line 52: @rafinskipg I think you had previously added this isOwner check? What does it mean to be an owner?

rafinskipg: @tyler17 since this list is also displayed when we access other peoples profile, to see their list of farms, i added the isowner to display only the actions (manage) if the page you are visiting is yours

rafinskipg:
Hey @tyler17 Gj, 

some comments:

I restored the other farm because initially the  idea is that you put DAI to farm Subdao tokens, and subdao tokens to farm MKR, so each subdao has 2 farms (that's initially, but idk if it has changed)

image


Second thing, is when i visit another profile : /address/ADDRESS i can see the farms of the user, that's where we use the isOwner , so we should keep the "isOwner" to not display the allowance.  

image


If i connect with another account that has not approved anything, or has no balances it also crashes, 

image

apps/webapp/src/modules/farms/components/FarmItem.tsx
tyler17: Aha, makes sense, thanks

rafinskipg (approved):

rafinskipg (approved):

apps/webapp/src/modules/farms/components/FarmItem.tsx
```diff
@@ -74,33 +93,39 @@ export function FarmItem({
...
-        {isOwner && (
-          <Button variant="outline">
-            {stakedBalance && stakedBalance > 0n
-              ? 'Manage position'
-              : `Deposit ${farm.stakeToken.symbol} to generate ${farm.rewardToken.symbol}`}
-          </Button>
-        )}
+        <FarmMetricsOverview farm={farm} ownerAddress={ownerAddress} stakedBalance={stakedBalance} />
+        <Flex sx={{ flexDirection: 'column' }}>
+          {!!rewardsBalance && (
```
torhector2, line 98: This won't show the rewards if they're 0n. Do we want that?

torhector2 (approved):
LGTM, just one question about showing rewards.

apps/webapp/src/modules/farms/components/FarmItem.tsx
torhector2: I guess it makes sense.

tyler17: good point, but yeah sounds good as is

b-pmcg: this comment made me think about showing "rewards earned to date," eg. sum of all past rewards claimed. Not sure if its possible to get that data from the contract without reading events though, but if possible it might be a nice feature to see.

b-pmcg: oh I guess we already have the query to calc that. Can add a ticket for it down the road after talking to Tiago

b-pmcg (approved):

PR #155, opened 2023-05-25
reset e2e (WIP)

rafinskipg:
I was trying to reset the e2e fork between tests. 

Problem: 
- the testnet is in another package, how dow we communicate cross package?
    - we have this cy.exec() that executes shell commands, it should be enough to put there the command i added (testnet:reset)
    - but.. where is the root path of the command executed? from inside the app folder? from the root of the monorepo?
    
    Another concern is , is it "okay?" to invoke a reset testnet from the webapp? those are 2 different apps and they shouldn't talk together like that. Or maybe it is?
 
 
 Anyway, this or more questions can only be answered with the experience and patiente of someone like @b-pmcg 
 
 i'm happy to pass you this task 😆 <3


b-pmcg:
> I was trying to reset the e2e fork between tests.
> 
> Problem:
> 
> * the testnet is in another package, how dow we communicate cross package?
>   
>   * we have this cy.exec() that executes shell commands, it should be enough to put there the command i added (testnet:reset)
>   * but.. where is the root path of the command executed? from inside the app folder? from the root of the monorepo?
>   
>   Another concern is , is it "okay?" to invoke a reset testnet from the webapp? those are 2 different apps and they shouldn't talk together like that. Or maybe it is?
> 
> Anyway, this or more questions can only be answered with the experience and patiente of someone like @b-pmcg
> 
> i'm happy to pass you this task 😆 <3

PR #158, opened 2023-05-29
Trust level: Data source tooltip -> Data source indicator

hernandoagf:
### What does this PR do?
- Change the `DataSourceTooltip` component for a new `DataSourceIndicator` component, consisting of a trust level colored button that triggers the `DataSourceModal`.
- Fix the error with data source indicators breaking out of the container
- Fix an error where opening a modal would freeze the site (modal remaining open but invisible).

image
image


adamgoth (approved):
I like this, nice improvement 💯

PR #159, opened 2023-05-31
Jstream widgets

b-pmcg:


torhector2:
We could move some components from the webapp to the components package so they can be used both for the webapp and the widgets package. If that makes sense 🤔 

Components like the WrapUnwrap, the Ssr modals, and farm manage components.

torhector2:
If we get the Wrap branch merged I think you could try with the WrapUnwrap component as it only needs to get the initial token as param to get it working. 
Ideally we could tell the chat "help me wrapping my MKR" and it'll show the widget with the MKR -> SYD selected. I'll check if there's any dependency first that could prevent this from working as expected. I'll let you know, in the meantime the SSR is prob the best option.

b-pmcg:
> UPDATE: Merged
> 
> If we get the Wrap branch merged I think you could try with the WrapUnwrap component as it only needs to get the initial token as param to get it working. Ideally we could tell the chat "help me wrapping my MKR" and it'll show the widget with the MKR -> SYD selected. I'll check if there's any dependency first that could prevent this from working as expected. I'll let you know, in the meantime the SSR is prob the best option.
> 
> You have an example on the Wrap.tsx with some optional props there, but the easier way to integrate it would be:
> 
> ```
> import { WrapUnwrap } from '@makerdao-dux/components';
> 
> <WrapUnwrap
>   initialToken="DAI"
> />
> ```

nice @torhector2, I think you're right swap is a good component to begin with since the designs match the latest designs that have more of a widget feel to it. I pulled in the latest from `main` with the swap components. I also added a generic widget wrapper that can convert an ethers provider to wagmi config if provided.

Got stuck on some unrelated things, I broke the build somehow, probably a tsconfig or vite.config issue
```
│ packages/widgets build: [commonjs--resolver] Missing "./dist/providers/jsonRpc" specifier in "wagmi" package
│ packages/widgets build: error during build:
│ packages/widgets build: Error: Missing "./dist/providers/jsonRpc" specifier in "wagmi" package
```

Will look more into it later

packages/widgets/src/components/ssr/index.tsx
```diff
@@ -0,0 +1,33 @@
...
+    </>
+  ) : (
+    <>Loading</>
+  );
+};
+
+// Exports the  widget with the wrapper, passes provider/client to wrapper
+export const SsrWidget = (props: SsrProps) => {
+  return (
+    <WidgetWrapper {...props}>
```
rafinskipg, line 29: are you sure about this approach of exporting the widngets with the wagmi provider already embedded?

I think the app that uses the widget should provide their own wrapper

```
<WagmiProvider>
  <Widget />
  </WagmiProvider>

```

Otherwise i think we will be embedding into these components all the wagmi provider code and when importing it might be a lot. I guess from the webapp we will be importing them too. 

I guess we could do 2 exports, one with the bundled wagmi , so , standalone import, and other with less dependencies.

PR #161, opened 2023-06-02
Farm pages bug fixes and chart style

hernandoagf:
### What does this PR do?
It fixes some bugs in farm pages and changes some styles in farm charts


torhector2 (approved):
Looking good, good job.

I think we need to decide what to display when there's no address. Not directly related with this PR but we don't have consistency over when to show the "Connet wallet" button. We need to talk with Tiago.

rafinskipg:
> Looking good, good job.
> 
> I think we need to decide what to display when there's no address. Not directly related with this PR but we don't have consistency over when to show the "Connet wallet" button. We need to talk with Tiago.

Agree, also all the loading states in the farm list should be updated with a better UI.

PR #162, opened 2023-06-05
Loaded

rafinskipg:
Theres a scenario where the app will connect to the default public RPC before using the configured by the user, which can be misleading or a potential vulnerability if the user is relying only on using local nodes.

We should make sure that the app is not rendering with the provider until localhost is accessed to load the configured providers


adamgoth (approved):

adamgoth (changes requested):
Deployed instance doesn't seem to load for me: https://metadao-frontend-l9wm1ikln-dux-core-unit.vercel.app/

PR #163, opened 2023-06-05
Detail layout

torhector2:
A new layout for detail pages.

Open for improvement but we can start working on detail pages based on this.
You can define in which side you want the fixed content, left or right, being right the default value.

https://github.com/makerdao-dux/metadao-frontend/assets/440694/843e0bd6-ab50-4440-9988-11b46873662e

### Right
image

### Left
image


packages/components/src/components/layout/DetailLayout.tsx
```diff
@@ -0,0 +1,49 @@
...
+  fixedContentSx,
+  fixedContentSide = 'right',
+  sx,
+  columnGap = 4,
+  top = 0
+}: DetailLayoutProps): React.ReactElement {
+  const isMobileLayout = useBreakpointValue({ base: true, lg: false });
+  const side = useMemo(
+    () => (
+      <Box as="aside" sx={{ ...fixedContentSx, position: 'sticky', top, height: '1px' }}>
```
adamgoth, line 26: The height needs to be 1px for this to work?

adamgoth (approved):
LGTM, just had one question

packages/components/src/components/layout/DetailLayout.tsx
torhector2: Yes, it need a height to work, if you put sth like 100vh it will have a weird dragging effect after reaching the bottom.

torhector2: it needs a height yes. I just changed it for 0 as it seems to work well.

adamgoth: nice, let's merge 🛳️

tyler17 (approved):

